PR #8, opened 2014-10-16
Added ability to fetch all available versions. Also there is method to get latest version.

sivolobov:
Now we can provide user with all available versions for symfony.
It includes some exotic version numbers like `2.5.0-BETA2` that is also available for download (http://get.symfony.com/Symfony_Standard_Vendors_2.5.0-BETA2.zip)


javiereguiluz:
Closing it because the installer already downloads the list of versions from symfony.com website and checks if the requested version is installable or not.

PR #28, opened 2014-10-20
Add global shorten command for symfony app/console

totophe:
Proposal of a global command shortened for symfony app/console that could be packaged with the symfony installer. It could be then evaluated as a more complete command than just a forwarding command in the future.

### DIFF
--- a/sf
+++ b/sf
@@ -2,9 +2,9 @@
 <?php
 
 	if (file_exists(getcwd() . DIRECTORY_SEPARATOR . 'app' . DIRECTORY_SEPARATOR . 'console')) {
-		chdir(getcwd());
-		array_shift($argv);		
-		passthru('app/console ' . implode(' ' , $argv));
+		array_shift($argv);
+		$escapedArgs = array_map(function($arg){ return escapeshellcmd($arg); }, $argv);
+		passthru('php app/console ' . implode(' ' , $escapedArgs));
 	} else {
 		echo "You must be in a symfony project to execute this command.\n";
 	}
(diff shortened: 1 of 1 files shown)


futurevision:
:+1:

weaverryan:
I don't think I like this - it's an extra layer of magic and means there'd be two ways of doing things that you'd find around the web. I could see questions like "what's the difference between `sf` and `app/console`" and also "command `sf` does not exist - how do I install this on my deploy server?"

Pierstoval:
I think like @weaverryan that it's a magic that is not totally needed. You could simply move the `./app/console` file in `./sf` and change the different include paths and it's done.
 
As the `app/console` is already a layer for the console, adding another layer is not much necessary :confused:

totophe:
I think the goal is more to have a global shortcut installed with the symfony-installer that detect if you are in a SF folder and shorten the app/console syntax. But maybe it should be integrated within the ```symfony``` installer and not be an extra command to maintain.

Pierstoval:
I still think it's app-specific, and should not be overriden "natively" by the installer nor by the standard edition itself :confused: 
 
I personally created some "./cs" shortcuts for *nix machines, and a "cs.cmd" for Windows, but it does not always work the expected way because of different shells and versions. For example, shells in Mac and Debian can differ in some small behaviors. But it's, as said, a *personal* choice (and it's only for one project).
 
I'm still :-1: for this, because it adds a new layer for something that is already a layer.

futurevision:
I have to agree with Pierstoval on this one...removed my :+1: for this PR.

javiereguiluz:
Although this proposal is not a bad idea at all, I must close it as *"won't fix"*. It seems that most people think that it has more drawbacks than benefits.

Thank you all for the discussion.

PR #63, opened 2014-11-30
Create the .gitignore file as part of the installation

unfunco:
The `.gitignore` file is not included as part of the installation.

I'm not sure whether this is the best method to accomplish this, since it potentially requires maintenance of the `$ignored` array, I'll gladly take suggestions...


javiereguiluz:
Closing it because the code of this PR was used in the more generic #103 PR which has just been merged. @unfunco thanks again for your help.

PR #89, opened 2014-12-22
test dependencies

xabbuh:
With this commit, Travis simply ensures that the installer dependencies do resolve properly. I will add "real" tests for the commands later on (you can merge this before and I can create a separate pull request later on).

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -16,7 +16,7 @@ use Distill\Exception\IO\Input\FileCorruptedException;
 use Distill\Exception\IO\Input\FileEmptyException;
 use Distill\Exception\IO\Output\TargetDirectoryNotWritableException;
 use Distill\Strategy\MinimumSize;
-use GuzzleHttp\Client;
+use GuzzleHttp\ClientInterface;
 use GuzzleHttp\Exception\ClientException;
 use GuzzleHttp\Message\Response;
 use GuzzleHttp\Subscriber\Progress\Progress;
@@ -36,6 +36,16 @@ use Symfony\Component\Filesystem\Filesystem;
 class NewCommand extends Command
 {
     /**
+     * @var Distill
+     */
+    private $distill;
+
+    /**
+     * @var ClientInterface
+     */
+    private $client;
+
+    /**
      * @var Filesystem
      */
     private $fs;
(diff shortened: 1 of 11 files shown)


javiereguiluz:
@xabbuh could you please remind me the status of this PR? I see that you worked hard creating unit tests and it's sad to see that they haven't been merged. I don't know what's left to do in this PR.

xabbuh:
@javiereguiluz Actually, I sadly forgot about this PR. I'll do a rebase and check if everything is still working. Thanks for the reminder.

xabbuh:
This can be reviewed again.

tests/Symfony/Installer/Tests/NewCommandTest.php
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * @var EmitterInterface|\PHPUnit_Framework_MockObject_MockObject
+     */
+    private $emitter;
+
+    private $projectsDir;
+
+    protected function setUp()
+    {
+        $this->chooser = $this->getMockBuilder('Distill\Chooser')
```
stof: mocking classes from other libraries don't ensure that our code works, because the other libraries might change its implementation and mocks will then be out of sync. You should not mock accross boundaries.

src/Symfony/Installer/NewCommand.php
```diff
@@ -210,9 +227,8 @@ private function download()
         $this->output->writeln("\n Downloading Symfony...");
 
         // decide which is the best compressed version to download
-        $distill = new Distill();
         $baseName = 'http://symfony.com/download?v=Symfony_Standard_Vendors_'.$this->version;
-        $symfonyArchiveFile = $distill
+        $symfonyArchiveFile = $this->distill
```
stof: given that ``Distill`` is a stateful object, using it as a reusable dependency looks wrong

src/Symfony/Installer/NewCommand.php
```diff
@@ -252,14 +268,13 @@ private function download()
             $progressBar->setProgress($downloaded);
         };
 
-        $client = new Client();
-        $client->getEmitter()->attach(new Progress(null, $downloadCallback));
+        $this->client->getEmitter()->attach(new Progress(null, $downloadCallback));
```
stof: the fact that you change the state of the client each time this method is called makes it unreliable if you use the client as a global dependency of the class:

- it will impact any later downloads for instance. 
- it will cause issues if the method is called several times (even if this case is unlikely as the logic is in a CLI command)

xabbuh:
I didn't have the time to dig deeper into setting things up properly. So I'm closing here. However, I will keep the branch on my fork. So anyone willing to finish this PR is welcome to just work on that branch. With #161 we also still have an issue reminding us that we really need tests for the installer.

PR #95, opened 2015-01-02
add LICENSE file

xabbuh:


xabbuh:
Actually, I first just wanted to update the year in the license, but then I noticed that there was no license file at all. Choosing the MIT license here was done on purpose given that this is already stated in the `composer.json` file and that it is consistent with the rest of the Symfony codebase.

LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2014-2015 Fabien Potencier
```
xabbuh: @javiereguiluz I'm not completely sure here since afaik you initially wrote the first version of the installer.

javiereguiluz: I think it's fine to keep Fabien :)

javiereguiluz:
Thanks Christian.

PR #97, opened 2015-01-04
Fix bug #symfony/12019

corsaroquad:
To fix this bug I have made an algorithm that sets the progress bar's width according to the columns of a terminal window.

The code uses external programs, "tput" for Linux/Unix systems and "mode" for Windows, to get the total columns of the window. In case of a terminal too wide, the code sets the width of the progress bar to the original value (60 chars). In case of a terminal too narrow, the code ignores the progress bar and displays only the file size and the percentage of the download. 

The bug was posted here: https://github.com/symfony/symfony/issues/13019

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -489,7 +489,7 @@ class NewCommand extends Command
     /**
      * Utility method to show the number of bytes in a readable format.
      *
-     * @param int     $bytes The number of bytes to format
+     * @param int $bytes The number of bytes to format
      *
      * @return string The human readable string of bytes (e.g. 4.32MB)
      */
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@corsaroquad thanks for providing a fix for this annoying problem.

However, I think that the fix should be submitted to Symfony Console component itself. If we fix the installer but let the Console error, every other console application will suffer from this bug. Or maybe I'm missing something else.

corsaroquad:
@javiereguiluz in reality it is not a bug, but a normal terminal behaviour. When we use carriage returns to erase/update/replace a printed string, but the terminal splits the string itself because the string has too many chars to fit in a line, the carriage return works only for the last line.

A possible solution for the Symfony Console component could be to do an helper to get these terminal's properties and save them into variables, but the user should be advised to use them if he wants to do a "responsive" printing. I wrote variables and not constants because the terminal size could change during the operation (for example someone that is having fun at resizing the terminal during the download of Symfony), but, personally, I suggest at this time to support *libnewt* and *libncurses* to do a nice terminal interface without doing a command overflow of *tput* and *mode* to refresh terminal properties.

However, I've searched on the PHP documentation and we could only use these external programs. **Warning**: they could change over time or there could be someone that hasn't that specific program installed (someone that hasn't *tput* installed on his linux terminal for example). 

This is a sad limitation of PHP, especially because it's useful knowing the size of the terminal window without using external libraries such as *libnewt* or *libncurses*.

stof:
@corsaroquad the Console Application already has some logic to determine the terminal width. Couldn't you rely on it instead of building a different implementation (which does not work on Windows)

javiereguiluz:
@corsaroquad I really appreciate that you took some time to help us fix/improve the installer. Sadly, I'm going to close this pull request without merging its changes. The reasons are:

  * This doesn't seem to be a bug of the installer, but of the Symfony Console component. That's why I think we should discuss about it in this related issue: https://github.com/symfony/symfony/issues/13019
  * As @stof said, the proposed solution uses its own logic to determine the terminal width, but the Console component already provides that and it should be reused.

PR #99, opened 2015-01-23
Allow to install Symfony when the directory exists but it's empty

### LINKED ISSUE #98: Allow to create project in current (Existing) directory just like Django does
Currently, Symfony installer version 0.5.0 not allow to create project in a current directory.
It would be nice this feature, just like [django-admin](https://docs.djangoproject.com/en/1.7/ref/django-admin/#startproject-projectname-destination) (from Django framwork) [does](https://docs.djangoproject.com/en/1.7/ref/django-admin/#startproject-projectname-destination):
```
~$ django-admin.py startproject  <projectname> [destination]
```
Where destination is a optional path to install in.  Using ‘.’ denotes the current working directory.
It would be:
```
~$ symfony new  <projectname> [destination] [version]
```
It would be too useful, for instance, when creating project from openshift where repository directory is created first.

javiereguiluz:
This fixes #98

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -614,7 +614,7 @@ class NewCommand extends Command
      * Checks whether the given directory is empty or not.
      *
      * @param  string  $dir the path of the directory to check
-     * @return bool         True if the directory is empty. False otherwise.
+     * @return bool
      */
     private function isEmptyDirectory($dir)
     {
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -97,7 +97,7 @@ class NewCommand extends Command
      */
     private function checkProjectName()
     {
-        if (is_dir($this->projectDir)) {
+        if (is_dir($this->projectDir) && !$this->isEmptyDirectory($this->projectDir)) {
             throw new \RuntimeException(sprintf(
                 "There is already a '%s' project in this directory (%s).\n".
                 "Change your project name or create it in another directory.",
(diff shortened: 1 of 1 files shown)


src/Symfony/Installer/NewCommand.php
```diff
@@ -609,4 +609,15 @@ private function getExecutedCommand()
 
         return sprintf('%s new %s %s', $executedCommand, $this->projectName, $version);
     }
+
+    /**
+     * Checks whether the given directory is empty or not.
+     *
+     * @param  string  $dir the path of the directory to check
+     * @return boolean      True if the directory is empty. False otherwise.
```
xabbuh: Should be `bool`.

src/Symfony/Installer/NewCommand.php
```diff
@@ -609,4 +609,15 @@ private function getExecutedCommand()
...
+     */
+    private function isEmptyDirectory($dir)
+    {
+        return 0 === count(glob($dir.'/*', GLOB_NOSORT));
```
xabbuh: IIRC `glob()` wouldn't return hidden files.

Pierstoval: Hmm, I checked and you're totally right. Probably would better use this code:
```php
return 2 === count(scandir($dir.'/'));
```
For the info, 2 is the number matching an empty dir where the only accessible elements are `.` and `..` , so count two :)

javiereguiluz: Thanks for reporting this error. This PR has been updated to use `scandir()` instead of `glob()`

Pierstoval:
:+1:  now :)

src/Symfony/Installer/NewCommand.php
```diff
@@ -609,4 +609,17 @@ private function getExecutedCommand()
 
         return sprintf('%s new %s %s', $executedCommand, $this->projectName, $version);
     }
+
+    /**
+     * Checks whether the given directory is empty or not.
+     *
+     * @param  string  $dir the path of the directory to check
+     * @return bool         True if the directory is empty. False otherwise.
```
stof: ``True if the directory is empty. False otherwise.`` could be removed. The name of the method already makes the meaning of the boolean clear

javiereguiluz: You are right. Removed that redundant explanation.

PR #103, opened 2015-02-10
Clean up Symfony-related stuff after creating the project

javiereguiluz:
This Pull Request "cleans" the installed project making the following changes:

  * Remove Symfony's `README.md` file and create a generic empty README.md
  * Remove all the Symfony's `UPGRADE*.md` files.
  * Remove the Symfony's `LICENSE.md` file.
  * Add the `.gitignore` file (and sort its entries)
  * Remove Symfony's `branch-alias` from `composer.json` file
  * Remove the `description` property from `composer.json` file
  * Put `proprietary` as the `license` in the `composer.json` file

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -366,11 +366,8 @@ class NewCommand extends Command
             $filesToRemove = array_merge($commonFiles, $upgradeFiles, $changelogFiles);
             $this->fs->remove($filesToRemove);
 
-            $now = new \DateTime('now');
-            $this->fs->dumpFile(
-                $this->projectDir.'/README.md',
-                sprintf("%s\n%s\n\nA Symfony project created on %s.\n", $this->projectName, str_repeat('=', strlen($this->projectName)), $now->format('F j, Y, g:i a'))
-            );
+            $readmeContents = sprintf("%s\n%s\n\nA Symfony project created on %s.\n", $this->projectName, str_repeat('=', strlen($this->projectName)), date('F j, Y, g:i a'));
+            $this->fs->dumpFile($this->projectDir.'/README.md', $readmeContents);
         } catch (\Exception $e) {
             // don't throw an exception in case any of the Symfony-related files cannot
             // be removed, because this is just an enhancement, not something mandatory
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -498,8 +498,14 @@ class NewCommand extends Command
 
         $contents['name'] = $this->generateComposerProjectName();
         $contents['license'] = 'proprietary';
-        unset($contents['description']);
-        unset($contents['extra']['branch-alias']);
+
+        if (isset($contents['description'])) {
+            unset($contents['description']);
+        }
+
+        if (isset($contents['extra']['branch-alias'])) {
+            unset($contents['extra']['branch-alias']);
+        }
 
         file_put_contents($filename, json_encode($contents, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES)."\n");
 
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -359,11 +359,11 @@ class NewCommand extends Command
         $this->fs->remove(dirname($this->compressedFilePath));
 
         try {
-            $commonFiles = array($this->projectDir.'/LICENSE', $this->projectDir.'/UPGRADE.md');
-            $upgradeFiles = glob($this->projectDir.'/UPGRADE-*.*.md');
-            $changelogFiles = glob($this->projectDir.'/CHANGELOG-*.*.md');
+            $licenseFile = array($this->projectDir.'/LICENSE');
+            $upgradeFiles = glob($this->projectDir.'/UPGRADE*.md');
+            $changelogFiles = glob($this->projectDir.'/CHANGELOG*.md');
 
-            $filesToRemove = array_merge($commonFiles, $upgradeFiles, $changelogFiles);
+            $filesToRemove = array_merge($licenseFile, $upgradeFiles, $changelogFiles);
             $this->fs->remove($filesToRemove);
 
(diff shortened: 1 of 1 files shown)


stof:
couldn't this been done when generating the archives ?

javiereguiluz:
@stof at first that was the idea, but I think it's easier and more convenient to do it with the installer. For example, this allows us to easily create a `README.md` file with the name of the project being created.

src/Symfony/Installer/NewCommand.php
```diff
@@ -356,6 +358,30 @@ private function cleanUp()
...
+            $this->fs->remove(array(
+                $this->projectDir.'/LICENSE',
+                $this->projectDir.'/UPGRADE.md',
+                $this->projectDir.'/UPGRADE-2.2.md',
+                $this->projectDir.'/UPGRADE-2.3.md',
+                $this->projectDir.'/UPGRADE-2.4.md',
+                $this->projectDir.'/UPGRADE-2.5.md',
+                $this->projectDir.'/UPGRADE-2.6.md',
+                $this->projectDir.'/UPGRADE-2.7.md',
+                $this->projectDir.'/UPGRADE-3.0.md',
```
stof: This list should be build dynamically with a glob, to remove any ``UPGRADE-*.md`` file. This would avoid having to update the installer for each minor version to add a new file. And what about the changelog

javiereguiluz:
I've refactored the code to use `glob()` to detect the `UPGRADE` and `CHANGELOG` files to remove.

romqin:
:+1:

src/Symfony/Installer/NewCommand.php
```diff
@@ -356,6 +358,22 @@ private function cleanUp()
     {
         $this->fs->remove(dirname($this->compressedFilePath));
 
+        try {
+            $commonFiles = array($this->projectDir.'/LICENSE', $this->projectDir.'/UPGRADE.md');
+            $upgradeFiles = glob($this->projectDir.'/UPGRADE-*.*.md');
```
xabbuh: Can't you use `UPGRADE*.md`?

src/Symfony/Installer/NewCommand.php
```diff
@@ -356,6 +358,22 @@ private function cleanUp()
...
+            $changelogFiles = glob($this->projectDir.'/CHANGELOG-*.*.md');
```
xabbuh: Afaics, there are no changelog files in the standard edition.

src/Symfony/Installer/NewCommand.php
```diff
@@ -476,12 +494,46 @@ private function updateComposerJson()
...
+            '/app/bootstrap.php.cache',
+            '/app/cache/*',
+            '!app/cache/.gitkeep',
+            '/app/config/parameters.yml',
+            '/app/logs/*',
+            '!app/logs/.gitkeep',
+            '/app/phpunit.xml',
+            '/bin/',
+            '/build/',
+            '/composer.phar',
```
xabbuh: What's the reason for the two entries above (`/build/` and `/composer.phar`)?

romqin: I guess because there are in the [original `.gitignore`](https://github.com/symfony/symfony-standard/blob/2.7/.gitignore)

javiereguiluz: @romqin is right, I copied the original `.gitignore` of the Standard Edition.

src/Symfony/Installer/NewCommand.php
javiereguiluz: You are right. There's no `CHANGELOG` files because the process that builds the ZIP/TGZ files removes them. But this single line of code will prevent us from future errors in case the ZIP/TGZ builder fails or changes.

javiereguiluz:
I made some minor changes to address your comments. This PR is ready for the final review before merging it. Thanks.

src/Symfony/Installer/NewCommand.php
xabbuh: I think we should remove them here. Otherwise, we could even simply keep the one from the standard edition.

src/Symfony/Installer/NewCommand.php
```diff
@@ -356,6 +358,22 @@ private function cleanUp()
     {
         $this->fs->remove(dirname($this->compressedFilePath));
 
+        try {
+            $licenseFile = array($this->projectDir.'/LICENSE');
+            $upgradeFiles = glob($this->projectDir.'/UPGRADE*.md');
+            $changelogFiles = glob($this->projectDir.'/CHANGELOG*.md');
```
xabbuh: @javiereguiluz What do you think about `*.md` then? You will probably not want to keep any markdown files from the SE at all.

src/Symfony/Installer/NewCommand.php
javiereguiluz: @xabbuh I'm afraid I don't understand you. Do you think we should remove this `createGitIgnore()` method from the installer?

xabbuh: I only mean the two entries `/build/` and `/composer.phar` (the latter could be argued if we expect people to install Composer locally).

src/Symfony/Installer/NewCommand.php
javiereguiluz: We want to maintain for example the `README.md` file (but changing its contents). Removing all the `.md` files seems too farsighted for me.

xabbuh: If present, a readme would we overriden at all right now, wouldn't it?

javiereguiluz: Honestly, I'm not comfortable removing any `*.md` file as suggested. I prefer to be careful and remove the files we know for certain that should be removed (CHANGELOG, UPGRADE, LICENSE).

src/Symfony/Installer/NewCommand.php
javiereguiluz: You are right. I've removed the strange `/build/` entry, but left the `composer.phar` entry.

weaverryan:
@javiereguiluz Great idea! So now, the user is basically downloading the SE (with vendors already) and then we're operating on it inside the installer. It means that pretty much every part is OS'ed and can be improved. Thanks :)

PR #105, opened 2015-02-20
Allow last LTS version to be installed.

94noni:
Hi, 

I just wanted to add the possibility to install the latest Symfony LTS version directly via the installer (new command).

Docs PR https://github.com/symfony/symfony-docs/pull/5032

Feedbacks welcome :)

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -137,6 +137,11 @@ class NewCommand extends Command
             return $this;
         }
 
+        // 'lts' is a special version name that refers to the current long term support version
+        if ('lts' === $this->version) {
+            return $this;
+        }
+
         // validate server syntax
         if (!preg_match('/^2\.\d(?:\.\d{1,2})?$/', $this->version)) {
             throw new \RuntimeException('The Symfony version should be 2.N or 2.N.M, where N = 0..9 and M = 0..99');
--- a/README.md
+++ b/README.md
@@ -45,7 +45,21 @@ $ symfony new my_project
 $ php symfony.phar new my_project
 ```
 
-**2. Start a new project based on a specific Symfony branch**
+**2. Start a new project with the latest Symfony LTS version**
+
+Execute the `new` command and provide the name of your project as the first
+argument and `lts` as the second argument. The installer will automatically
+select the most recent LTS version available:
+
+```bash
+# Linux, Mac OS X
+$ symfony new my_project lts
+
+# Windows
+$ php symfony.phar new my_project lts
+```
+
+**3. Start a new project based on a specific Symfony branch**
 
 Execute the `new` command and provide the name of your project as the first
 argument and the branch number as the second argument. The installer will
@@ -59,7 +73,7 @@ $ symfony new my_project 2.3
 $ php symfony.phar new my_project 2.3
 ```
 
-**3. Start a new project based on a specific Symfony version**
+**4. Start a new project based on a specific Symfony version**
 
 Execute the `new` command and provide the name of your project as the first
 argument and the exact Symfony version as the second argument:
--- a/src/Symfony/Installer/AboutCommand.php
+++ b/src/Symfony/Installer/AboutCommand.php
@@ -54,6 +54,11 @@ class AboutCommand extends Command
 
    <comment>%s new blog</comment>
 
+ To create a new project called <info>blog</info> in the current directory using
+ the <info>long term support version</info> of Symfony, execute the following command:
+
+   <comment>%s new blog lts</comment>
+
  To base your project on a <info>specific Symfony branch</info>, append the branch
  number at the end of the command:
 


src/Symfony/Installer/NewCommand.php
```diff
@@ -133,7 +139,22 @@ private function checkSymfonyVersionIsInstallable()
     {
         // 'latest' is a special version name that refers to the latest stable version
         // available at the moment of installing Symfony
-        if ('latest' === $this->version) {
+        if (self::VERSION_LATEST === $this->version) {
+            return $this;
+        }
+
+        // Get the Symfony versions list (http://symfony.com/versions.json)
+        $versionsList = $this->getVersionsList();
```
stof: this is no loading the version list in cases where it is not needed (passing an exact version), which is wrong. The HTTP request should be done only when needed

src/Symfony/Installer/NewCommand.php
```diff
@@ -36,6 +36,12 @@
 class NewCommand extends Command
 {
     /**
+     * Constants.
+     */
+    const VERSION_LATEST = 'latest';
+    const VERSION_LTS = 'lts';
```
stof: I don't think these constants make much sense here.

src/Symfony/Installer/NewCommand.php
94noni: Yes of course, going to fetch+cache them

src/Symfony/Installer/NewCommand.php
94noni: The idea was to be able to deal with symfony download

https://github.com/symfony/symfony-installer/blob/master/src/Symfony/Installer/NewCommand.php#L218

http://symfony.com/download?v=Symfony_Standard_Vendors_latest => _can_ be resolved
http://symfony.com/download?v=Symfony_Standard_Vendors_lts => _can not_ be resolved (exception here https://github.com/symfony/symfony-installer/blob/master/src/Symfony/Installer/NewCommand.php#L270)

src/Symfony/Installer/NewCommand.php
xabbuh: Well, this is still only necessary if the user doesn't pass a version number in the `2.M.N` format (to either fetch the latest patch version, the LTS version or the latest stable version).

src/Symfony/Installer/NewCommand.php
```diff
@@ -133,7 +139,22 @@ private function checkSymfonyVersionIsInstallable()
...
+        }
+
+        // Get the Symfony versions list (http://symfony.com/versions.json)
+        $versionsList = $this->getVersionsList();
+
+        // 'lts' is a special version name that refers to the current long term support version
+        if (self::VERSION_LTS === $this->version) {
+            if ($versionsList && isset($versionsList[$this->version])) {
+                // Get the latest patch of the lts version
+                $this->version = $versionsList[$this->version];
```
xabbuh: Does this really select the latest LTS?

src/Symfony/Installer/NewCommand.php
stof: I don't see how dealing with this is related to the fact of defining these constants

src/Symfony/Installer/NewCommand.php
stof: @94noni what you should do is move this call inside the ``if (self::VERSION_LTS === $this->version) {``  conditional (and add it again in the place where you removed it). The call won't be done twice anyway because the JSON gives you the exact version for the LTS, not the branch

Pierstoval: Unless `symfony.org` plans to propose a proper "standard" download link for the LTS version, this behavior seems correct to me.

src/Symfony/Installer/NewCommand.php
stof: ``$versionList['lts']`` is the exact version yes. Whether it is always the latest is a matter of the server-side logic

src/Symfony/Installer/NewCommand.php
```diff
@@ -44,6 +50,7 @@ class NewCommand extends Command
     private $version;
     private $compressedFilePath;
     private $requirementsErrors = array();
+    private $versionsList = null;
```
stof: I don't think you need to make it stateful here. It is needed only once in each case anywya

src/Symfony/Installer/NewCommand.php
```diff
@@ -682,4 +699,17 @@ private function isEmptyDirectory($dir)
         // scandir() returns '.'  and '..'  for an empty dir
         return 2 === count(scandir($dir.'/'));
     }
+
+    /**
+     * Set the Symfony versions list (http://symfony.com/versions.json)
+     *
+     * @return $this
+     */
+    private function setVersionsList()
```
stof: the name is weird. It is not a setter

src/Symfony/Installer/NewCommand.php
94noni: I thought first for readability purpose

src/Symfony/Installer/NewCommand.php
Pierstoval: There is a `lts` key in the `http://symfony.com/versions.json` file. As the `$this->version` contains `lts` when reaching this line, it retrieves the correct element.

94noni: based on the json, yes

src/Symfony/Installer/NewCommand.php
94noni: ok so lets remove this + setter function

javiereguiluz:
If you want, we can make `lts` version name to work on symfony.com. We already do this for `latest` version name and it would be really easy to do. This way we could simplify a lot the code of this PR. What do you think?

94noni:
@javiereguiluz exactly what i asked for @stof 

> can you please just explain me this:
> http://symfony.com/download?v=Symfony_Standard_Vendors_latest => can be resolved
> http://symfony.com/download?v=Symfony_Standard_Vendors_lts => can not be resolved 
> it will be easier (and no more query call for json for the lts) if it is resolved correctly :)

Pierstoval:
This could reduce this PR to two lines of code, I think it's worth the cost :smiley:

javiereguiluz:
OK. The `lts` version is already supported in symfony.com. But we must wait for the related pull request to get approved and deployed in the production website. I'll keep you posted.

xabbuh:
@javiereguiluz Will this be the latest LTS once Symfony 2.7 is released?

javiereguiluz:
@xabbuh yes. It will always point to the most recent LTS version, whichever it is.

stof:
@94noni can you update the PR so that it simply handle ``lts`` in the same way than ``latest`` (which should indeed be a 2 lines patch), even if the archive is not yet in production (it only means we have to wait before merging, but it does not prevent to have the patch ready)

94noni:
Thank all of you, waiting your go to change this (and PR on the related symfony documentation)

stof:
Can you revert the other changes being done here to really keep it a 2-line patch (well, maybe 4 because of formatting)

94noni:
@stof PR finalized, much cleaner with the symfony.com lts version supported now

stof:
:+1:

94noni:
Just forget the read me. Updated now

javiereguiluz:
The special `lts` version name is already available in production:

```
http://symfony.com/download?v=Symfony_Standard_Vendors_lts.zip
http://symfony.com/download?v=Symfony_Standard_Vendors_lts.tgz
```

stof:
:+1: then

94noni:
Cool :)
And is it ok for the https://github.com/symfony/symfony-docs/pull/5032 as well?

javiereguiluz:
Thanks @94noni! It took me a lot of time to merge this, but it's finally done :)

PR #111, opened 2015-02-28
Typo in comment fixed

King2500:
I think this is probably a typo and was meant to be "semver" (for semantic versioning).

Right?

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -137,7 +137,7 @@ class NewCommand extends Command
             return $this;
         }
 
-        // validate server syntax
+        // validate semver syntax
         if (!preg_match('/^2\.\d(?:\.\d{1,2})?$/', $this->version)) {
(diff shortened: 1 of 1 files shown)


King2500:
Ping @javiereguiluz

src/Symfony/Installer/NewCommand.php
```diff
@@ -137,7 +137,7 @@ private function checkSymfonyVersionIsInstallable()
             return $this;
         }
 
-        // validate server syntax
+        // validate semver syntax
```
xabbuh: I would actually change this to "validate version syntax" or "validate version pattern".

javiereguiluz:
Thank you @King2500.

src/Symfony/Installer/NewCommand.php
javiereguiluz: @xabbuh your proposal is easier to understand ... but any developer looking at the source code of this command probably knows what `semver` is :)

PR #113, opened 2015-03-01
Custom installer

aprezcuba24:
Coding Standard and Common Typos


xabbuh:
@aprezcuba24 Can you pleas explain a bit your idea for this pull request?

Pierstoval:
It's linked to #112 , but it's not a good idea IMO. The installer is here to install Symfony. If you want to install something else, you might not need the Symfony installer to do it.

aprezcuba24:
First of all, sorry for any mistake, I do not speak English and it is the first time I collaborate on a project and I don't know how to do it.
In Symfony1 it was possible to implement own installers to generate projects with this command.

```bash
   $ php /path/to/symfony generate:project --installer=/somewhere/my_installer.php
```

With this functionality the developer could define additional actions, such as intall plugins, run commands, etc.
This installer for Symfony2, this possibility does not exist. One of the utilities that could have is that the user defining it as: other vendors to be installed, create services, set parameters, generate the database and run fixtures, etc.
I put the following example, suppose you want that when the installation is done as follows:
  - more elements add to it in the .gitignore
  - New vendors will add to it
  - The vendors upgrade to the latest version.

For example:

```php
namespace CustomInstaller;

use Symfony\Install\Installer;
use Symfony\Component\Process\Process;

class CustomInstaller extends Installer
{
    protected $gitIgnoreEntries = array(
        '/app/bootstrap.php.cache',
        '/app/cache/*',
        '!app/cache/.gitkeep',
        '/app/config/parameters.yml',
        '/app/logs/*',
        '!app/logs/.gitkeep',
        '/app/phpunit.xml',
        '/bin/',
        '/composer.phar',
        '/vendor/',
        '/web/bundles/',

        //Add oder gitignore
        '.idea',
        '/web/uploads'
    );

    public function getMethodsSequence()
    {
        return array_merge(parent::getMethodsSequence(), [
            'addVendors',
            'updateVendors',
        ]);
    }

    public function addVendors()
    {
        $this->runProcess('composer require knplabs/knp-paginator-bundle --prefer-dist');
    }

    public function updateVendors()
    {
        $this->runProcess('php /home/util/composer.phar update --prefer-dist');
    }

    protected function runProcess($command)
    {
        $this->output->writeln("\n ".$command);
        $process = new Process($command, $this->projectDir, null, null, null);
        $process->run(function ($type, $buffer){
            if (Process::ERR === $type) {
                $this->output->writeln(sprintf('     <error>%s</error>', $buffer));
            } else {
                $this->output->writeln(sprintf('     <info>%s</info>', $buffer));
            }
        });

        if (!$process->isSuccessful()) {
            throw new \RuntimeException($process->getErrorOutput());
        }
    }
}

return new CustomInstaller();
```

then run the command
```bash
symfony new blog --installer=/path/to/CustomInstaller.php
```

Pierstoval:
It seems that what you want is multiple Symfony apps, all with their own settings.
 
For this, you may choose between these 2 choices:

* Install 2 apps in 2 different directories. The basic, simplest way to do it.
* Install a single app, require all your vendors in it, and create one `Kernel` for each one, and maybe one other `console` file and `app`/`app_dev` file for each app. Tricky but functional.
 
Allowing the user to setup a custom installer is kind of heavy, and will not probably be used that much by users. 
 
Plus, for your example, if you need a custom installer just to require and update vendors, it seems rather useless, as `composer` does it very well on its own with short `composer install` and `composer update`.

javiereguiluz:
@aprezcuba24 thanks for the time you took to prototype this feature and to help us improve the installer.

Sadly, I have to close this pull request as *won't fix*. The reason is that this feature is currently way out of the scope of the installer. It's too advanced/complex for what we want the installer to be.

As a matter of fact, we consider the Symfony Installer finished. In a few months, we may revisit this assumption, but from now on, we'll only focus on fixing bugs.

aprezcuba24:
thanks, Only say, the "CustomInstaller" class is a simple example. With this funcionality the developers may do all they want.

Issue #108, opened 2015-02-27
Self-update should work when the file is writable but not its folder


stof:
Currently, it fails when trying to rename the file. This forces to use sudo to run the update in ``/usr/local/bin``.

The self-update logic available in Composer works fine for this case (the command used here is based on an old version of the composer logic, not the current one).

javiereguiluz:
Fixed in #114

PR #114, opened 2015-03-02
SelfUpdate command now is more robust and allows to rollback

javiereguiluz:
This is just a proposal for improving the SelfUpdate command and fixing #108 (if you agree with it, I'll add things like PHPdoc).

### DIFF
--- a/src/Symfony/Installer/SelfUpdateCommand.php
+++ b/src/Symfony/Installer/SelfUpdateCommand.php
@@ -75,7 +75,7 @@ class SelfUpdateCommand extends Command
 
         $this->remoteInstallerFile = 'http://symfony.com/installer';
         $this->currentInstallerFile = realpath($_SERVER['argv'][0]) ?: $_SERVER['argv'][0];
-        $this->tempDir = is_writable(dirname($this->currentInstallerFile)) ? dirname($this->currentInstallerFile) : sys_get_temp_dir();
+        $this->tempDir = sys_get_temp_dir();
         $this->currentInstallerBackupFile = basename($this->currentInstallerFile, '.phar').'-backup.phar';
         $this->newInstallerFile = $this->tempDir.'/'.basename($this->currentInstallerFile, '.phar').'-temp.phar';
         $this->restorePreviousInstaller = false;
@@ -93,6 +93,7 @@ class SelfUpdateCommand extends Command
                 ->checkNewVersionIsValid()
                 ->backupCurrentVersion()
                 ->replaceCurrentVersionbyNewVersion()
+                ->cleanUp()
             ;
         } catch (\Exception $e) {
             $this->rollback();
@@ -166,6 +167,11 @@ class SelfUpdateCommand extends Command
         $this->fs->copy($this->newInstallerFile, $this->currentInstallerFile, true);
 
         return $this;
+    }
+
+    private function cleanUp()
+    {
+        $this->fs->remove(array($this->currentInstallerBackupFile, $this->newInstallerFile));
     }
 
     private function rollback()
--- a/src/Symfony/Installer/SelfUpdateCommand.php
+++ b/src/Symfony/Installer/SelfUpdateCommand.php
@@ -97,7 +97,6 @@ class SelfUpdateCommand extends Command
             ;
         } catch (\Exception $e) {
             $this->rollback();
-            throw $e;
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/Symfony/Installer/SelfUpdateCommand.php
```diff
@@ -45,71 +55,106 @@ public function isEnabled()
...
-            $fs->remove($tempFilename);
-            $output->writeln(sprintf('<error>The downloaded file is corrupted (%s).</error>', $e->getMessage()));
-            $output->writeln('<error>Please re-run the self-update command to try again.</error>');
-
-            return 1;
-        }
+    private function rollback()
+    {
+        $this->fs->remove($this->tempInstallerFilePath);
+        $this->fs->rename($this->previousInstallerFile, $this->localInstallerFile);
```
stof: this renaming should be done only if the roolback is done *after* the backup of the old version. Otherwise it could rallback to an even older version.

and using ``Filesystem::rename`` works only when the parent folder is writable IIRC, meaning you would break during the rollback when the command failed early.

src/Symfony/Installer/SelfUpdateCommand.php
```diff
@@ -45,71 +55,106 @@ public function isEnabled()
...
-        try {
-            $tempFilename = $tempDir.'/'.basename($localFilename, '.phar').'-temp.phar';
+        return $this;
+    }
 
-            $fs->copy($remoteFilename, $tempFilename);
-            $fs->chmod($tempFilename, 0777 & ~umask());
+    private function backupOldVersion()
+    {
+        $this->fs->rename($this->localInstallerFile, $this->previousInstallerFile);
```
stof: renaming te file means that you will then be forced to write to the folder instead of writing to the file, thus not solving #108

Pierstoval: Would you recommend then to copy it and remove it when the command terminates?

stof: this is what composer is doing (except it does not even remove it later because it allows ``composer self-update --rollback``)

javiereguiluz:
I've made a lot of changes as suggested. Please review it again and tell me what you think. Thanks!

src/Symfony/Installer/SelfUpdateCommand.php
```diff
@@ -45,71 +62,114 @@ public function isEnabled()
...
+        $this->restorePreviousInstaller = true;
 
-            $fs->rename($tempFilename, $localFilename, true);
+        return $this;
+    }
 
-            $output->writeln('<info>Symfony Installer was successfully updated.</info>');
+    private function replaceOldVersionbyNewVersion()
+    {
+        $fs->copy($this->newInstallerFile, $this->currentInstallerFile, true);
```
Pierstoval: Shouldn't this be `$this->fs` instead of `$fs`?

src/Symfony/Installer/SelfUpdateCommand.php
```diff
@@ -27,6 +27,23 @@
  */
 class SelfUpdateCommand extends Command
 {
+    /** @var Filesystem */
+    private $fs;
+    /** @var OutputInterface */
+    private $output;
+    private $tempDir;
+    // URL where the latest installer version can be downloaded
```
stof: this should be phpdoc given that it documents the property

src/Symfony/Installer/SelfUpdateCommand.php
```diff
@@ -45,71 +62,114 @@ public function isEnabled()
...
-            $tempFilename = $tempDir.'/'.basename($localFilename, '.phar').'-temp.phar';
+        return $this;
+    }
 
-            $fs->copy($remoteFilename, $tempFilename);
-            $fs->chmod($tempFilename, 0777 & ~umask());
+    private function backupOldVersion()
+    {
+        $this->fs->copy($this->currentInstallerFile, $this->currentInstallerBackupFile, true);
+        $this->fs->copy($this->newInstallerFile, $this->currentInstallerFile, true);
```
stof: this second command is not about doing a backup. It is the job of ``replaceOldVersionbyNewVersion``

javiereguiluz:
I've updated this PR again and with the changes introduced in a3e31ab, I think this is finished and ready to be merged. I've tested it and everything seems to work as expected.

PR #118, opened 2015-03-19
Added a new "symfony demo" command

javiereguiluz:
This is a WIP and early release of the new `symfony demo` command. The general idea around this command is the following:

  * Soon the `AcmeDemoBundle` will be gone forever.
  * The Symfony Standard Edition (and therefore, the `symfony new` command) will always give you a clean and empty Symfony application.
  * We'll publish a complete Symfony demo application that combines what we did for the Best Practices book (and which wasn't released) and the current AcmeDemoBundle. This will be a real application using `AppBundle` and all the best practices. No more Acme or Demo.
  * If you are new to Symfony and want to test it ... install the Symfony Installer in 1 minute and execute the `symfony demo` command. In 10 seconds you'll have a full-featured Symfony application ready to test.

And here are some comments regarding this command:

  * For now it doesn't download anything because the `Symfony_Demo` package doesn't exist yet.
  * This command **must** be as simple as possible (remember that it will be used by newcomers that just want to test-drive Symfony quickly and easily). That's why we don't allow to select the Symfony version (it will always be the most recent one). And we also don't allow to select the directory where the demo project is installed (it will be installed in `getcwd().'symfony_demo/'`)

And here it is the `demo` command in action:

symfony_demo_command

### DIFF
--- a/src/Symfony/Installer/DownloadCommand.php
+++ b/src/Symfony/Installer/DownloadCommand.php
@@ -1,0 +1,345 @@
+<?php
+
+/*
+ * This file is part of the Symfony Installer package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Installer;
+
+use Distill\Distill;
+use Distill\Exception\IO\Input\FileCorruptedException;
+use Distill\Exception\IO\Input\FileEmptyException;
+use Distill\Exception\IO\Output\TargetDirectoryNotWritableException;
+use Distill\Strategy\MinimumSize;
+use GuzzleHttp\Client;
+use GuzzleHttp\Exception\ClientException;
+use GuzzleHttp\Message\Response;
+use GuzzleHttp\Subscriber\Progress\Progress;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\ProgressBar;
+
+/**
+ * Abstract command used by commands which download and extract compressed Symfony files.
+ *
+ * @author Christophe Coevoet <stof@notk.org>
+ * @author Javier Eguiluz <javier.eguiluz@gmail.com>
+ */
+abstract class DownloadCommand extends Command
+{
+    /**
+     * Chooses the best compressed file format to download (ZIP or TGZ) depending upon the
+     * available operating system uncompressing commands and the enabled PHP extensions
+     * and it downloads the file.
+     *
+     * @return Command
+     *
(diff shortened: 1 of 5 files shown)


Pierstoval:
:+1: 
 
I'm gonna use this command in my trainings once it becomes available :+1:

xabbuh:
Just a quick suggestion: Instead of copying a lot of code from the `new` command, we should probably some parts to a dedicated layer and reuse it in both commands. As a benefit, this would probably also ease testing them quite a lot (I still need to wrap up my pull request though).

javiereguiluz:
@xabbuh what would you think if I make the `DemoCommand` inherit from the `NewCommand`?

xabbuh:
In this case, an abstract base command class would probably be better.

javiereguiluz:
@xabbuh I've updated this PR to introduce the common abstract method.

src/Symfony/Installer/DownloadCommand.php
```diff
@@ -0,0 +1,345 @@
...
+            } else {
+                throw new \RuntimeException(sprintf(
+                    "There was an error downloading %s from symfony.com server:\n%s",
+                    $this->getDownloadedFileName(),
+                    $e->getMessage()
+                ));
+            }
+        }
+
+        $this->fs->dumpFile($this->downloadedFilePath, $response->getBody());
```
stof: ``$this->fs`` should be moved to the DownloadCommand as it is used here

src/Symfony/Installer/DownloadCommand.php
```diff
@@ -0,0 +1,345 @@
...
+     * available operating system uncompressing commands and the enabled PHP extensions
+     * and it downloads the file.
+     *
+     * @return Command
+     *
+     * @throws \RuntimeException if the Symfony archive could not be downloaded
+     */
+    protected function download()
+    {
+        $this->output->writeln(sprintf("\n Downloading %s...\n", $this->getDownloadedFileName()));
```
stof: ``$this->output`` should be moved to this class

stof: same for any other property used here btw (look at Insight which can probably warn you IIRC)

src/Symfony/Installer/DownloadCommand.php
```diff
@@ -0,0 +1,345 @@
...
+     */
+    protected function getDownloadedFileName()
+    {
+        $commandName = $this->getName();
+
+        if ('new' === $commandName) {
+            return 'Symfony';
+        }
+
+        if ('demo' === $commandName) {
```
stof: using an abstract method implemented in each command would be cleaner IMO

PR #120, opened 2015-03-26
fix handling of missing patch version

xabbuh:
When the user tried to install the Symfony Standard Edition with just
specifying the minor version (like `symfony new blog 2.3`), the
installer failed for two reasons:

* The `use` statement for the `Client` class was missing when it was
  used to download the list of available versions from symfony.com.
* The URL from where the archive should be downloaded was build before
  the patch version was evaluated. Therefore, the archive of the actual
  Symfony version to be installed could not be found (it was tried to
  download an archive for `2.3` while the archive actually had to be
  referenced for `2.3.26`).

This fixes #119.

PR #121, opened 2015-03-26
Improved the Download abstract command

javiereguiluz:
This PR addresses the comments made by @stof in #118 and tries to improve the emergency solution provided by @xabbuh in #120.

### DIFF
--- a/src/Symfony/Installer/DemoCommand.php
+++ b/src/Symfony/Installer/DemoCommand.php
@@ -13,7 +13,6 @@ namespace Symfony\Installer;
 
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Filesystem\Filesystem;
 
 /**
  * This command creates a full-featured Symfony demo application.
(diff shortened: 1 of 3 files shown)


src/Symfony/Installer/DownloadCommand.php
```diff
@@ -32,6 +35,24 @@
 abstract class DownloadCommand extends Command
 {
     /**
+      * Returns the type of the downloaded application in a human readable format.
+      * It's mainly used to display readable error messages.
+      * @return string
+      */
```
xabbuh: Looks like you need one space in front of each *.

src/Symfony/Installer/DemoCommand.php
```diff
@@ -26,7 +25,6 @@ class DemoCommand extends DownloadCommand
     protected $fs;
```
xabbuh: This property should be moved to the `DownloadCommand` class.

javiereguiluz: You are right. I moved the initialization of the variables but not their declaration. Done. Thanks.

PR #122, opened 2015-03-26
Update README.md

JulienItard:
Update Linux installation step


stof:
the advantage of the previous instruction is that it does not run sudo as root

JulienItard:
Yes i know but i think it should be the same installation step here and official website.

javiereguiluz:
@JulienItard thanks for this pull request. Indeed it's a good idea to have the same installation instructions everywhere. Thanks!

PR #125, opened 2015-03-29
set PATH environment variable for Windows

sstok:
Windows uses Path instead of PATH which causes a notice for the about and download command.

### DIFF
--- a/symfony
+++ b/symfony
@@ -20,6 +20,11 @@ require file_exists(__DIR__.'/vendor/autoload.php')
 
 $appVersion = '1.0.3-DEV';
 
+// Windows uses Path instead of PATH
+if (!isset($_SERVER['PATH']) && isset($_SERVER['Path'])) {
+    $_SERVER['PATH'] = $_SERVER['Path'];
+}
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@sstok thanks for this improvement. It looks totally safe to merge it, but since I'm not a Windows user, I prefer to wait until any Symfony developer with Windows experience (e.g. @stof or @WouterJ) confirms it. Thanks for your patience.

Pierstoval:
I personally use Windows 7 as my dev environments, and it has never caused me any problem... :confused:

Pierstoval:
I personally use Windows 7 as my dev environments, and it has never caused me any problem... :confused:

WouterJ:
@javiereguiluz I can confirm that this PR is correct (@sstok and me actually created this PR when helping a windows users with this error)

@Pierstoval the path env is only used by the command in the help message (if you run `symfony`) or if there was an error during the download. That might explain why you didn't see this message.

javiereguiluz:
@WouterJ thanks for reviewing this change and thanks again to @sstok for improving the installer

Issue #129, opened 2015-03-31
Installation not possible in the existing directory that can not be removed.


gander:
`~/bin/symfony-installer/symfony new ~/workspace/`

`Fatal error: Uncaught exception 'RuntimeException' with message 'Could not delete /home/codio/workspace: rmdir(/home/codio/workspace): Device or resource busy' in /home/codio/bin/symfony-installer/vendor/raulfraile/distill/src/Extractor/Util/Filesystem.php:177`

Pierstoval:
Have you tried to go to your home directory and try to execute `symfony new workspace` ?

gander:
`rm: cannot remove 'workspace': Device or resource busy`

This directory can not be removed. This is blocked by the system, and the intended effect. Why Symfony Installer must delete this folder?

Pierstoval:
Is the directory empty ?

gander:
>codio:~$ pwd 
>/home/codio 
>codio:~$ ls -al /home/codio/workspace/ 
>total 4 
>drwxr-xr-x  2 codio codio    6 Mar 31 12:10 . 
>drwxr-xr-x 24 codio codio 4096 Mar 31 12:08 .. 
>codio:~$ rm -fr /home/codio/workspace/ 
>rm: cannot remove `/home/codio/workspace': Device or resource busy 

But why you are focusing on the fact that I can not delete this folder? This is not a problem! Let's focus on that to improve the installer.

Pierstoval:
It's not normal that the installer tries to delete this directory :confused:

gander:
>codio:~$ ls -al workspace 
>total 4 
>drwxr-xr-x  2 codio codio    6 Mar 31 12:58 . 
>drwxr-xr-x 24 codio codio 4096 Mar 31 12:08 .. 
>codio:~$ symfony -vvv new workspace 
> 
> Downloading Symfony... 
> Preparing project... 
> 
>  [RuntimeException] 
>  Symfony can't be installed because the downloaded package is corrupted 
>  or because the installer doesn't have enough permissions to uncompress and 
>  rename the package contents. 
>  To solve this issue, check the permissions of the /home/codio directory and 
>  try executing this command again: 
>  symfony new workspace 
> 
>Exception trace: 
> () at phar:///home/codio/bin/symfony/src/Symfony/Installer/DownloadCommand.php:162 
[...]

PR #131, opened 2015-03-31
Add previous exception for some 'unknown errors'

Pierstoval:
At the moment, we don't have that many clues when people can't install Symfony and when we can't guess what is the problem with the backtrace.

This PR allows debugging easier, especially when Distill does throw an exception.
 
This is related to #129


javiereguiluz:
@Pierstoval :+1: I like it. To me it's a safe and useful merge, but I'll wait to see if there are more code reviews. Thanks.

WouterJ:
:+1:

xabbuh:
:+1:

javiereguiluz:
Thanks @Pierstoval.

PR #132, opened 2015-04-01
Added support for using proxy servers

javiereguiluz:
I've tested it successfully with some free public proxy servers, but we should definitely test it with "real" proxy servers.

### DIFF
--- a/src/Symfony/Installer/DownloadCommand.php
+++ b/src/Symfony/Installer/DownloadCommand.php
@@ -42,6 +42,7 @@ abstract class DownloadCommand extends Command
     /**
      * Returns the type of the downloaded application in a human readable format.
      * It's mainly used to display readable error messages.
+     *
      * @return string
      */
     abstract protected function getDownloadedApplicationType();
@@ -112,7 +113,7 @@ abstract class DownloadCommand extends Command
             $progressBar->setProgress($downloaded);
         };
 
-        $client = new Client();
+        $client = $this->getGuzzleClient();
         $client->getEmitter()->attach(new Progress(null, $downloadCallback));
 
         // store the file in a temporary hidden directory with a random name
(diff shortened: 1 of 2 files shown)


src/Symfony/Installer/DownloadCommand.php
```diff
@@ -150,6 +151,25 @@ protected function download()
     }
 
     /**
+     * Returns the Guzzle client configured according to the system environment
+     * (e.g. it takes into account whether it should use a proxy server or not).
+     *
+     * @return Client
+     */
+    protected function getGuzzleClient()
```
xabbuh: Is there a reason not to make this private?

xabbuh:
The `NewCommand` class also uses the Guzzle client to download the list of available versions.

src/Symfony/Installer/DownloadCommand.php
javiereguiluz: Nice catch! Fixed. Thanks.

javiereguiluz:
Nice catch! Fixed. Thanks.

src/Symfony/Installer/DownloadCommand.php
sstok: You can't call private methods in subclasses :)

xabbuh: It wasn't used in another class the time I wrote the comment. :D

javiereguiluz: You both are right! Thanks for reviewing the code.

javiereguiluz:
I'm going to merge this now and release a new installer version. The code looks "safe enough" to not break anything if you are not using a proxy. And it also looks like we're going to solve the problem at least for some proxy users.

PR #138, opened 2015-04-06
Added a additional check if the posix extension is loaded

jdecoster:
Added a check to prevent a thrown exception when the posix extension is not loaded in php.ini


sstok:
:+1:

javiereguiluz:
@jdecoster thanks for fixing this issue!

PR #139, opened 2015-04-06
Don't create README.md on cleanup

xelaris:
In case of an exception, e.g. when typing an invalid version number, the `cleanUp` method gets called. Since the `README.md` file was dumped there, the target directory wasn't empty on the next try, so it wasn't possible to re-run the command without manually removing the file.


javiereguiluz:
@xelaris thanks for fixing this issue and for improving the user experience with this change.

PR #140, opened 2015-04-08
Fix error due to bad file path when installer is located at drive root o...

leto3087:
...n Windows

### DIFF
--- a/src/Symfony/Installer/DownloadCommand.php
+++ b/src/Symfony/Installer/DownloadCommand.php
@@ -117,7 +117,7 @@ abstract class DownloadCommand extends Command
         $client->getEmitter()->attach(new Progress(null, $downloadCallback));
 
         // store the file in a temporary hidden directory with a random name
-        $this->downloadedFilePath = getcwd().DIRECTORY_SEPARATOR.'.'.uniqid(time()).DIRECTORY_SEPARATOR.'symfony.'.pathinfo($symfonyArchiveFile, PATHINFO_EXTENSION);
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@leto3087 thanks for creating this pull request.

Since I don't use Windows, before merging it I'll wait for @WouterJ @sstok and other Windows users to review it. Thanks.

WouterJ:
Seems perfectly valid to me.

Btw, can't we simply use `\\` instead of DIRECTORY_SEPARATOR here? As it's only a problem on Windows, we know it's going to be `\\`.

xabbuh:
The current working directory could end with a backslash on other systems. Even if that's not very likely, we should not introduce any possible issues here by hardcoding `\\` instead of the `DIRECTORY_SEPARATOR` constant.

javiereguiluz:
Everybody seems happy with this PR and the code looks correct, so I'm going to merge it so we can test it in real systems. Thanks @leto3087 for fixing this issue!

bluppfisk:
Problem still exists, Windows 10, using command prompt. Works fine if I use Bash on Windows (Linux Subsystem), but not with the command prompt. Mind, I use C:/wamp64/www/ as my projects directory to allow web-serving through Apache.

```
    5.09 MB/5.09 MB ============================================================  100%

 Preparing project...

  [RuntimeException]
  Symfony can't be installed because the downloaded package is corrupted
  or because the installer doesn't have enough permissions to uncompress and
  rename the package contents.
  To solve this issue, check the permissions of the C:\wamp64\www directory and
  try executing this command again:
  symfony new projectname

  [RuntimeException]
  C:\wamp64\www\projectname does not exist and could not be created.
```

Issue #143, opened 2015-04-14
Installation of dev/beta versions


xelaris:
I wonder if `http://get.symfony.com/Symfony_Standard_Vendors_2.7.0-dev.tgz` should be available or the installer should not try to fetch it. When calling `symfony new blog 2.7`, version 2.7 is resolved to `2.7.0-dev`, since it is provided so by http://symfony.com/versions.json, so the installer tries to fetch the file and stops with "The selected version (2.7.0-dev) cannot be installed because it does not exist.".
Maybe the installer should respect the list of `installable` versions, provided by `versions.json`, or forbid "branch-versions" in general?
Further, if it should be possible to install 2.7 with the installer, I would expect to get `2.7.0-beta1` now.

I guess it is not intended to enable installation of dev/beta versions? Then, what about providing a message like:

> You are trying to install an unstable version. Please use composer to install such versions:
>
> $ composer create-project symfony/framework-standard-edition blog (at)beta
>

Pierstoval:
I think it's not intended to enable installation of unstable(like dev or beta) versions.
 
To use SF 2.7, just install 2.6 with `symfony new project 2.6` and change your composer.json file to fetch `~2.7@beta` version.

javiereguiluz:
@xelaris thanks for opening this issue. I think it's an interesting discussion and I have no clear answer.

My opinion is that we should only install stable versions ... but I'm totally open to hear other people's opinions about this.

And I agree with your proposal of showing a message if the user tries to install a `dev` version. The problem is, how can we detect correctly that it's a `dev` version? If the user writes exactly `2.7.0-dev`, we can detect that; but if the user writes `2.7`, I don't know how can we detect that.

Pierstoval:
Technically, a `dev` version is any version tag matching `x.y.z` under the `dev` key in the [versions.json](http://symfony.com/versions.json) file, isn't it?

stof:
@javiereguiluz if he writes 2.7, you are asking symfony.com for the exact version, right ? This means you can detect the dev version after that too.

xelaris:
Like @stof said, after resolving 2.7 via the `versions.json` it is possible to check the version for a 'dev' and/or 'beta' postfix and provide such a message. I have opened a related PR.

PR #144, opened 2015-04-14
Detect dev/beta versions and provide appropriate instruction

xelaris:
see #143

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -182,6 +182,15 @@ class NewCommand extends DownloadCommand
             ));
         }
 
+        if (preg_match('/^(\d\.\d)(\.\d{1,2})?-(dev|beta)\d*$/i', $this->version, $matches)) {
+            throw new \RuntimeException(sprintf(
+                "You are trying to install an unstable version (%s).\n".
+                "Please use composer to install such versions:\n\n".
+                'composer create-project symfony/framework-standard-edition %s "%s.*" --stability=%s',
+                $this->version, $this->projectDir, $matches[1], strtolower($matches[3])
+            ));
+        }
+
         return $this;
     }
 


xelaris:
Fabbot criticizes the double quotes, but they are used everywhere in the file. How should I handle this?

stof:
@xelaris ignore it outside your own code for now, until we fix it everywhere. The reason is that until recently, fabbot was not able to fix quotes. It was implemented only recently.

xelaris:
Thanks @stof. I fixed the unnecessary double quotes related to my change.

javiereguiluz:
:+1: the code looks good to me. Thanks @xelaris.

src/Symfony/Installer/NewCommand.php
```diff
@@ -182,6 +182,15 @@ protected function checkSymfonyVersionIsInstallable()
             ));
         }
 
+        if (preg_match('/^(\d\.\d).*-(dev|beta)\d*$/', $this->version, $matches)) {
```
xabbuh: I think the regular expression should be `/^(\d\.\d).*-(dev|beta\d*)$/` instead. Usually there will be no number behind the `-dev` suffix. And what about making the match case-insensitive to support version numbers like `2.7-BETA`?

xabbuh: Anyway, what is the reason for the `.*` part?

xelaris:
@xabbuh You are right about the number behind `-dev` suffix, but if I change it to `(dev|beta\d*)` the `betaX` must be converted to `beta` to use it for the stability argument for composer. Since I can't see any drawbacks in also matching e.g. `2.7-dev1`, I wonder if it is worth to do this. What do you think?

The `.*` part was there because the actual version is `2.7.0-dev` when looking up `2.7` and since the patch level isn't relevant IMO, I added it to be more tolerant, though `(.\d{1,2})?` is more accurate here, so I changed it.

Making it case-insensitive seems to be a good idea. Please have a look at the updated PR.

94noni:
IMO if a new comer tries to install a dev/beta version (that he may had read in a blog or forum) with the symfony installer and see the following message with composer, he will be so confused
Why not simply write that such unstable version are not installable for the moment?

Pierstoval:
I think it's good to propose to install the dev versions, but only on the
current `dev` branch, because it allows developers to automatize the
Symfony installation even with the `dev` branch.
A newcomer will often install the last version, because he'd have saw it on
symfony.com.
I don't think that someone would tell his followers to write something such
as `symfony new project 2.7-dev` in a blog or a forum presenting Symfony...

2015-04-16 11:57 GMT+02:00 Antoine Makdessi <notifications@github.com>:

94noni:
Ok, so why talking about composer?
why not directly add the ``dev`` option that is resolved like the ``latest`` or ``lts`` in the json file? or let the installer resolve a version (e.g. 2.7 to 2.7.0-dev) and warn the developer that the version being installed is a unstable one?

Pierstoval:
You're right, we should not talk about composer as it is not used when retrieving the correct Symfony version.

stof:
@Pierstoval symfony.com does not generate pre-built archives for branches (which would require regenerating them on each commit)

Pierstoval:
So there is no "currently possible" way to install Symfony 2.7 with the installer?

javiereguiluz:
I propose you the following. Instead of outputting the Composer installation instructions, which may confuse some users, we could link to the new documentation article that explains how to use and install an inestable Symfony version.

This article isn't published in the docs yet (and it hasn't even been accepted yet) but here it is: https://github.com/symfony/symfony-docs/pull/5186

94noni:
If there is no way to generate a package for dev/beta versions, this is far better I think

javiereguiluz:
@94noni I'm afraid there is no way to generate those packages. As pointed by @stof, that would require us to generate a new package whenever a new pull request is merged.

In any case, using a inestable version is a "niche" thing with a simple alternative solution, as explained in the new doc. That's why I thing there is no need to complicate things.

Pierstoval:
:+1: for the docs

xabbuh:
@javiereguiluz Would it be much overhead to build them for beta versions too? Making that possible would make it easier for people to test them and hopefully find issues before the final release.

stof:
it could be possible for beta versions IMO. these are releases after all

javiereguiluz:
To sum up the above discussion, what about doing the following:

  * If the user types in the name of a released beta version, the installer installs that beta version (we already have packages for beta versions: e.g. `http://get.symfony.com/Symfony_Standard_Vendors_2.7.0-BETA1.zip`).
  * For `dev` versions, we don't provide packages or any way to install them, but we show the user the doc link with more information.

Pierstoval:
`symfony install beta-test 2.7.0-beta1` should be good IMO :+1:

javiereguiluz:
@Pierstoval beware that URL is case sensitive:

```
// it doesn't work
$ symfony new my_project 2.7.0-beta1

// it works
$ symfony new my_project 2.7.0-BETA1
```

Pierstoval:
Thanks for pointing that, it's kind of important, we should then choose to `strtoupper` the argument then, or indicate explicitly that the beta is in uppercase :smiley:

94noni:
So 
```symfony new my_project 2.7```
```symfony new my_project dev```
are resolved and downloaded now?

**Comment by javiereguiluz** [comment]
@94noni nope :(

This already works (today):

```
$ symfony new my_project
$ symfony new my_project 2.6
$ symfony new my_project 2.7.0-BETA1
```

This doesn't work (today):

```
$ symfony new my_project 2.7
$ symfony new my_project 2.7.0-DEV
$ symfony new my_project 2.7.0-beta1
```

94noni:
Is a PR welcome for integrating dev and 2.7 args?

xabbuh:
@94noni I don't see a way to resolve this for `dev` (there is no package that could be downloaded). For `2.7`, I can think of one solution: Detect that there is a beta release available and ask the user if they want to install it. We should not do this automatically imho since people usually want to install a stable version of the framework.

94noni:
@xabbuh according to http://symfony.com/versions.json, "dev" => "2.7.0-BETA1" so if there is a zip ready on the server, it will be downloaded, no?
for 2.7 (or any dev/beta versions), I am ok with the proposition to ask before continuing the install of such versions

Pierstoval:
Can any branch present in the [versions.json](http://symfony.com/versions.json) file can be considered as "installable" except `dev` ?
 
If yes, there's not much to do in the installer to allow them to be installed, just need to add a check that matches "BETA" to ensure that the "probably uninstallable" branch is effectively installable, except `dev` branches (that are not released, so there is not distant zipball available for `dev` versions).

xelaris:
@javiereguiluz `symfony new my_project 2.7` is possible now, but installs `2.7.0-BETA1` without any information or warning. `symfony new my_project 2.7.0-BETA1` doesn't work, since the version has to match `/^2\.\d(?:\.\d{1,2})?$/`

@94noni Although `dev` points to `2.7.0-BETA1` yet and this would be an installable package, it could also point to `2.8.0-dev` which woudn't an installable package.

@Pierstoval e.g. the 2.8 branch, listed by `versions.json` now, can't be considered as installable, since it is a dev version only and there is no release yet.
It is already possible to call `symfony new my_project 2.8` now, but it ends in a HTTP request to `http://symfony.com/download?v=Symfony_Standard_Vendors_2.8.0-dev` which isn't available and won't be available in future. The command stops with a `RuntimeException` in this case.

I think, regardless of the listed branches in `versions.json`, the actual version must be checked for `-dev` or `-BETAX` postfix, to determine the stability. If there is no postfix, it is a stable release and can be installed like before. If it is a beta release, a warning or question should be presented, as suggested by @xabbuh. If there is no release (dev version) the link to the doc should be presented instead of the composer command as suggested by @javiereguiluz.

To enable a user to specify beta versions explicitly as an argument, the validation has to be changed and the command should ensure the `-betaX` is converted to uppercase.
I see another issue, once a `2.7.0-BETA2` comes out. Then there is no way to determine if `2.7.0-BETA1` is a valid/installable version, since there is no occurrence in `versions.json`, isn't it?

Pierstoval:
The `-dev` versions cannot be downloaded, as they are not released, so there is no package for it. And it will probably never be downloadable, as it would need the team to generate a zipball for every commit/merge in the branch...

xelaris:
@Pierstoval If your last comment is related to mine... I don't said dev versions should be downloaded, do I? Actually this PR was created to prevent the installer to try to download such versions.

javiereguiluz:
@xelaris thanks for proposing this feature and for sending this pull request ... and thank you all for the discussion.

I'm closing this in favor of #185, which hopefully addresses most of the raised concerns. Thanks!

Issue #147, opened 2015-04-21
Error when running "symfony self-update"


ZCJ:
// updating Symfony Installer to 1.1.1 version

There was an error while updating the installer.
The previous Symfony Installer version has been restored.

xabbuh:
@ZCJ Can you please run it again with the `-vvv` option and check if you get a more detailed error message?

ZCJ:
@xabbuh I have tried it, but no further information .. screen shot attached

![screen shot 2015-04-21 at 12 01 39 pm](https://cloud.githubusercontent.com/assets/1475845/7247782/3bbcba1a-e81e-11e4-957a-350e38d7271b.png)

javiereguiluz:
@ZCJ have you tried with `sudo symfony self-update`?

ZCJ:
sudo did it ... my bad ...

javiereguiluz:
@ZCJ I told you about `sudo` because this problem also caused a WTF!? moment to me. The problem is that the error message is not helpful at all. I'm going to prepare a pull request to improve it. For now, I'll close this issue as "fixed". Thanks.

ZCJ:
Exactly.. A better error msg will help .. Thank you

dsteplight:
Running "sudo symfony selfupdate" worked for me. It updated to the latest version which then allowed to successfully installed Symfony 3.2 .

PR #148, opened 2015-04-21
Improved error message when "self-update" command fails

javiereguiluz:
This change improves the error messages to avoid issues like #147

### DIFF
--- a/src/Symfony/Installer/SelfUpdateCommand.php
+++ b/src/Symfony/Installer/SelfUpdateCommand.php
@@ -20,7 +20,7 @@ use Symfony\Component\Filesystem\Exception\IOException;
 /**
  * This command is inspired by the self-update command included
  * in the PHP-CS-Fixer library. See:
- * https://github.com/fabpot/PHP-CS-Fixer/blob/master/Symfony/CS/Console/Command/SelfUpdateCommand.php
+ * https://github.com/fabpot/PHP-CS-Fixer/blob/master/Symfony/CS/Console/Command/SelfUpdateCommand.php.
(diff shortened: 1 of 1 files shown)


src/Symfony/Installer/SelfUpdateCommand.php
```diff
@@ -15,11 +15,12 @@
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Filesystem\Exception\IOException;
 
 /**
  * This command is inspired by the self-update command included
  * in the PHP-CS-Fixer library. See:
- * https://github.com/fabpot/PHP-CS-Fixer/blob/master/Symfony/CS/Console/Command/SelfUpdateCommand.php
+ * https://github.com/fabpot/PHP-CS-Fixer/blob/master/Symfony/CS/Console/Command/SelfUpdateCommand.php.
```
Pierstoval: Is a trailing `.` really needed here ?

stof: the proper fix would be to move this to the long description of the class instead

javiereguiluz: It's pretty awful, but required by the latest PHP-CS fixer version. Otherwise, fabbot will complain :cry:

Pierstoval: And why not use a phpdoc `@link` tag?

PR #152, opened 2015-04-29
Use SSL everywhere

javiereguiluz:
`symfony.com` enabled SSL recently, so we should switch to `https://` everywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,14 +12,14 @@ This step is only needed the first time you use the installer:
 ### Linux and Mac OS X
 
 ```bash
-$ sudo curl -LsS http://symfony.com/installer -o /usr/local/bin/symfony
+$ sudo curl -LsS https://symfony.com/installer -o /usr/local/bin/symfony
 $ sudo chmod a+x /usr/local/bin/symfony
 ```
 
 ### Windows
 
 ```bash
-c:\> php -r "readfile('http://symfony.com/installer');" > symfony
+c:\> php -r "readfile('https://symfony.com/installer');" > symfony
 ```
 
 Move the downloaded `symfony` file to your projects directory and execute
--- a/src/Symfony/Installer/DemoCommand.php
+++ b/src/Symfony/Installer/DemoCommand.php
@@ -129,6 +129,6 @@ class DemoCommand extends DownloadCommand
 
     protected function getRemoteFileUrl()
     {
-        return 'http://symfony.com/download?v=Symfony_Demo';
+        return 'https://symfony.com/download?v=Symfony_Demo';
     }
 }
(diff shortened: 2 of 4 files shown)


weaverryan:
Great news! Now, will this cause problems with some users (those that need to use the fopen solution)? If anyone has a pretty basic Windows setup and can try this, it'd be great to at least know what error some people will get. My impression is that fopen and SSL depends on your PHP setup (but I'm not totally sure).

Thanks!

Pierstoval:
I don't remember well (it's been a long time since the last time I had time to work at home :disappointed: ) but I think openssl is not enabled by default on windows, so obviously we need either a fallback solution (non-https) or a clear warning message for the end-users to warn them to activate openssl in their php config.

javiereguiluz:
The simple solution would be to use `https://` for Linux/Mac and `http://` for Windows.

Pierstoval:
Maybe you could check first whether openssl is enabled and fall back on `http` if it's not AND if the user is on Windows?

stof:
@javiereguiluz please use SSL on any system when openssl is enabled.

WouterJ:
I can't remember having any issues using Composer on my windows PC the first time. However, downloading some binaries (5.6.0 and 5.5.x) from http://windows.php.net and installing Wamp, both turn out to not have openssl enabled.

However, just like @stof and @Pierstoval, I advise to check if ssl is available and always use it if it is and otherwise, use http and output a warning with some information on how to enable openssl.

weaverryan:
I also think checking for ssl is a good idea - and I think we should be able to test this on non-windows machines just by disabling the openssl extension.

But what about the readfile line for windows users? We'd have to show them the https version with a note to change to http if that doesn't work - that's not a great experience. Instead, we could have them physically download the file via their browser.

javiereguiluz:
Closing it because the server infrastructure is not ready yet for SSL. If you try to execute this command, you'll get the following error:

```
$ sudo curl -LsS https://symfony.com/installer -o /usr/local/bin/symfony
curl: (7) Failed to connect to get.sensiolabs.org port 443: Connection refused
```

WouterJ:
Does this also apply to all other URLs, or just the download? Otherwise, I propose to still recommend downloading the installer using http (as https doesn't work yet), but already use https for the Symfony downloads.

javiereguiluz:
I prefer to wait a few days until SSL is widely supported across all servers and services and then we can make the switch for everything.

WouterJ:
"a few days" has become "a few months", is there any update on getting an SSL certificate for get.sensiolabs.com? This still is an important issue imo

xabbuh:
@javiereguiluz What is the status here?

javiereguiluz:
I'm afraid that SensioLabs's infrastructure is not ready yet to use SSL for this. I don't know when it will be ready. // cc @lyrixx

lyrixx:
If it's only about downloading .phar, we could put cloudfront in front of s3.
**but** this will imply some weirdness with cache

PR #155, opened 2015-05-04
Added support for a specific directory in demo command

### LINKED ISSUE #154: Add a directory argument to the demo command
Hi guys!

Small detail - I was just using the new `symfony demo` command (yay!) and I immediately wrote `symfony demo foo_bar_baz` - to make sure it didn't run over any existing directories. But, this argument doesn't exist.

I think it makes sense to have this argument (but not the version argument) - it's consistent with `symfony new`.

Thanks!

lyrixx:
fixes #154

### DIFF
--- a/src/Symfony/Installer/DemoCommand.php
+++ b/src/Symfony/Installer/DemoCommand.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Installer;
 
+use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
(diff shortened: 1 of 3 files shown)


src/Symfony/Installer/DemoCommand.php
```diff
@@ -38,22 +38,29 @@ protected function initialize(InputInterface $input, OutputInterface $output)
     {
         parent::initialize($input, $output);
 
-        $this->projectDir = getcwd();
+        if ('symfony_demo' === $input->getArgument('directory')) {
```
xabbuh: I would do this only if the argument has not been given at all, but not depending on the vaue if given.

lyrixx: indeed; fixed.

javiereguiluz:
Thanks @lyrixx.

src/Symfony/Installer/DownloadCommand.php
```diff
@@ -150,6 +152,19 @@ protected function download()
         return $this;
     }
 
+    protected function checkProjectName()
```
stof: you should remove the method from NewCommand too

lyrixx: https://github.com/symfony/symfony-installer/pull/157

PR #156, opened 2015-05-05
fix uppercase usernames

Ma27:
uppercase usernames will cause issues when submitting applications to packagist.

This bugfix transforms uppercase usernames like Ma27 to ma27 and all uppercase letters will be transformed to dashes

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -380,17 +380,36 @@ class NewCommand extends DownloadCommand
         $name = strtolower($name);
 
         if (!empty($_SERVER['USERNAME'])) {
-            $name = $_SERVER['USERNAME'].'/'.$name;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Ma27 thanks for preparing this PR. I have a question about it: should we apply the new method to the entire package name instead of just to the vendor part? Thanks.

Ma27:
well actually you are right.
I would remove the following lines https://github.com/Ma27/symfony-installer/blob/patch-1/src/Symfony/Installer/NewCommand.php#L379-L380 and apply the fixComposerVendorName to the whole package name and open a new PR

stof:
@javiereguiluz we should generate lowercase package names. Composer will treat them as lowercase anyway.

javiereguiluz:
With the latest changes I think this is ready to merge it because now we generate lowercased and dash-separated package names. Thanks @Ma27 for working on this!

PR #166, opened 2015-06-09
silence wrong exception

### LINKED ISSUE #163: Installer requires Intl
```
$ php symfony.phar new foo

 Downloading Symfony...

    4.92 MB/4.92 MB ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  100%

 Preparing project...

  [Symfony\Component\Intl\Exception\MethodArgumentValueNotImplementedException]
  The Symfony\Component\Intl\Collator\Collator::__construct() method's argument $locale
  value 'fr_FR' behavior is not implemented. Only the locale "en" is supported.  Please install
  the "intl" extension for full localization capabilities.

new directory [version]
```
http://symfony.com/doc/current/reference/requirements.html — this page states that Symfony has optional dependency on Intl, why does installer require it?

Symfony Installer has version 1.1.3.

xabbuh:
This fixes #163 by silencing the mistakenly thrown exception.

### DIFF
--- a/src/Symfony/Installer/DownloadCommand.php
+++ b/src/Symfony/Installer/DownloadCommand.php
@@ -25,6 +25,7 @@ use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Intl\Exception\MethodArgumentValueNotImplementedException;
 use Symfony\Installer\Exception\AbortException;
 
 /**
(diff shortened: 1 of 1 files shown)


weaverryan:
we should open an issue or a PR on the SE to remind fabpot to update the SE to the latest edition of the distribution bundle to fix this. I'd hate to have 2.7.1 released without it.

I'm +1 for this - we might as well fix it here to solve issues. We can remove it later when it's only affecting old versions. For a tool that's meant to help people, I think not merging this would be a little counter productive. So, yes for me :)

javiereguiluz:
This is really nice @xabbuh. Thanks a lot for working on this. And I agree with @weaverryan, we should fix this problem upstream in Symfony as soon as possible.

xabbuh:
> we should open an issue or a PR on the SE to remind fabpot to update the SE to the latest edition of the distribution bundle to fix this.

Doesn't this happen automatically when the archive is built?

weaverryan:
@xabbuh I actually don't know if a full `composer update` is part of updating the SE, or if fabpot just updates the composer.lock to a new `symfony/symfony` version. Anyways, he *did* update it before the 2.7.1 release today:

PR #169, opened 2015-06-17
Updated Windows install command so it is compatible with PowerShell

futurevision:
This should fix the garbage output in PowerShell on Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ $ sudo chmod a+x /usr/local/bin/symfony
 ### Windows
 
 ```bash
-c:\> php -r "readfile('http://symfony.com/installer');" > symfony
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,7 +19,7 @@ $ sudo chmod a+x /usr/local/bin/symfony
 ### Windows
 
 ```bash
-c:\> php -r "readfile('http://symfony.com/installer');" > symfony
+c:\> php -r "file_put_contents('symfony', file_get_contents('http://symfony.com/installer'));"
```
javiereguiluz: Opening URLs with the `file_get_contents()` function requires the `allow_url_fopen` INI setting to be `true`. Anyone know if that's the case for most Windows users?

javiereguiluz:
@futurevision thanks for providing this solution. It seems to be what we were looking for. Thanks!

PR #173, opened 2015-06-24
Update the composer.lock file to avoid out-of-sync errors

### LINKED ISSUE #172: The new project contains a mismatched composer.lock
```
$ symfony new new-project ; cd new-project/ ; composer install
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. Run update to update them.
```

I think it would be less confusing / make a less broken impression if it was at least in sync with itself for newly generated projects.

javiereguiluz:
This fixes #172

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -424,8 +424,7 @@ class NewCommand extends DownloadCommand
         $composerFileContents = file_get_contents($this->projectDir.'/composer.json');
         $lockFileContents = json_decode(file_get_contents($this->projectDir.'/composer.lock'), true);
 
-        $hash = md5($composerFileContents);
-        $lockFileContents['hash'] = $hash;
+        $lockFileContents['hash'] = md5($composerFileContents);
 
         file_put_contents($this->projectDir.'/composer.lock', json_encode($lockFileContents, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES)."\n");
     }
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -336,6 +336,8 @@ class NewCommand extends DownloadCommand
 
         file_put_contents($filename, json_encode($contents, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES)."\n");
 
+        $this->syncComposerLockFile();
+
         return $this;
     }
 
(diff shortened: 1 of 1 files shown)


src/Symfony/Installer/NewCommand.php
```diff
@@ -412,4 +414,19 @@ protected function getRemoteFileUrl()
...
+    /**
+     * Updates the 'hash' value stored in composer.lock to avoid out-of-sync
+     * problems when the composer.json file contents are changed.
+     */
+    private function syncComposerLockFile()
+    {
+        $composerFileContents = file_get_contents($this->projectDir.'/composer.json');
+        $lockFileContents = json_decode(file_get_contents($this->projectDir.'/composer.lock'), true);
+
+        $hash = md5($composerFileContents);
```
Pierstoval: Where did you find this behavior? In Composer itself?

src/Symfony/Installer/NewCommand.php
```diff
@@ -412,4 +414,19 @@ protected function getRemoteFileUrl()
...
+        $lockFileContents['hash'] = $hash;
+
+        file_put_contents($this->projectDir.'/composer.lock', json_encode($lockFileContents, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES)."\n");
```
Pierstoval: This will throw an error for PHP5.3 because the two constants do not exist. You might better use the `192` value which is the combination of these two constants.

javiereguiluz: The installer doesn't work with PHP 5.3. We prevent that here: https://github.com/symfony/symfony-installer/blob/master/symfony#L4

src/Symfony/Installer/NewCommand.php
javiereguiluz: I found it here: https://github.com/composer/composer/blob/69210d5bc130f8cc9f96f99582a041254d7b9833/src/Composer/Factory.php#L309  But to be sure I made an `md5` of several `composer.json` files and they matched the values stored in `composer.lock`.

src/Symfony/Installer/NewCommand.php
Pierstoval: Have you tested the behavior to be sure the hash is correct?

javiereguiluz: Sure! Here it is the summarized console output:

```bash
$ ./symfony new before

 Downloading Symfony...
 [...]

$ cd before/
$ composer install
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Warning: The lock file is not up to date with the latest changes in composer.json.
         You may be getting outdated dependencies. Run update to update them.
Nothing to install or update
Generating autoload files
[...]

$ cd ..
$ ./symfony new after

 Downloading Symfony...
 [...]

$ cd after
$ composer install
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Nothing to install or update
Generating autoload files
[...]
```

Pierstoval: Awesome then! :+1:

stof:
@javiereguiluz reimplementing the composer hashing here has a drawback though: the logic will become much more complex once composer improves its hashing by hashing only the relevant informations (which would require you to update a second hash for the new API).
But anyway, I guess we can live with this solution for now. Nobody is currently working on improving the hash computation, so it won't be there soon AFAICT. So :+1: from me

javiereguiluz:
@stof I agree with you. Let's merge this for now and if someday Composer changes the hashing algorithm, we'll match their implementation or drop this feature, whichever is easier for us.

PR #176, opened 2015-07-13
Check if the installer has enough permissions *before* downloading the file

### LINKED ISSUE #175: Check write permission *before* downloading
I think the installer should check for write permission **before** downloading (or at least reused the last downloaded one after permissions are fixed)

    symfony new my-project

    Downloading Symfony...

    4.95 MB/4.95 MB ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  100%

    [Symfony\Component\Filesystem\Exception\IOException]
    Failed to create "/var/www/.143660761855a0e482d53bd": mkdir(): Permission denied.

After that, if I fix the permissions and run `symfony new my-project` again, then Symfony is re-downloaded.

Although it's not a problem for me, it could be for those who have a slow connection.

javiereguiluz:
This fixes #175

### DIFF
--- a/src/Symfony/Installer/DemoCommand.php
+++ b/src/Symfony/Installer/DemoCommand.php
@@ -62,6 +62,7 @@ class DemoCommand extends DownloadCommand
         try {
             $this
                 ->checkProjectName()
+                ->checkPermissions()
                 ->download()
                 ->extract()
                 ->cleanUp()
--- a/src/Symfony/Installer/DownloadCommand.php
+++ b/src/Symfony/Installer/DownloadCommand.php
@@ -349,7 +349,7 @@ abstract class DownloadCommand extends Command
             return;
         }
 
-        $errorMessage  = wordwrap($requirement->getTestMessage(), $lineSize - 3, PHP_EOL.'   ').PHP_EOL;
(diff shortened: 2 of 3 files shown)


xabbuh:
Shouldn't we rather fix this in Distill? Otherwise we start to make assumptions on the library's internal behaviour.

javiereguiluz:
@xabbuh thanks for your comment. To be honest, I'm not sure what to say about this. Making this check in the installer is easy and pragmatic ... but passing this responsibility upstream is even more pragmatic. What should we do?

weaverryan:
:+1: and then we can open an issue/PR upstream. I *do* think that there should be some way (e.g. a try-catch) of knowing if the unzip fails and - preferably - why.

javiereguiluz:
@weaverryan most of the errors related to unzipping the contents are handled by the great work done by @raulfraile in https://github.com/symfony/symfony-installer/blob/master/src/Symfony/Installer/DownloadCommand.php#L206-234  but we also want to detect some errors *before* downloading the ZIP file.

weaverryan:
Oh yea, that's already really great stuff from Distill. So, I'm +1 for this one, simple check beforehand as an earlier warning.

Issue #92, opened 2014-12-24
Download/install fails on windows+vagrant


heitorvrb:
Windows 8.1, vagrant running an ubuntu server 14.04. Downloaded the installer, tried to run `symfony new testproject` and got the following error:

```
[GuzzleHttp\Exception\RequestException]
  Error creating resource. [url] http://symfony.com/download?v=Symfony_Standa
  rd_Vendors_latest.tgz [type] 2 [message] fopen(http://symfony.com/download?
  v=Symfony_Standard_Vendors_latest.tgz): failed to open stream: HTTP request
   failed!  [file] phar:///usr/local/bin/symfony/vendor/guzzlehttp/guzzle/src
  /Adapter/StreamAdapter.php [line] 367
```

Then I tried `wget http://symfony.com/download?v=Symfony_Standard_Vendors_latest.tgz` just to see what would happen:

```
HTTP request sent, awaiting response... 200 OK
Length: 5132534 (4.9M) [application/x-tar]
download?v=Symfony_Standard_Vendors_latest.tgz: Protocol error

Cannot write to ‘download?v=Symfony_Standard_Vendors_latest.tgz’ (Protocol error).
```

Searched around about that error, and it seems to be related to the `?` and the fact that I'm running inside a windows filesystem, where that character is not allowed in filenames. Did the same wget command with the `--restrict-file-names=windows` option and it downloaded normally, although it saved the file as `download@v=Symfony_Standard_Vendors_latest.tgz`:

```
 wget http://symfony.com/download?v=Symfony_Standard_Vendors_latest.tgz --restrict-file-names=windows
--2014-12-24 01:31:25--  http://symfony.com/download?v=Symfony_Standard_Vendors_latest.tgz
Resolving symfony.com (symfony.com)... 79.125.119.210
Connecting to symfony.com (symfony.com)|79.125.119.210|:80... connected.
HTTP request sent, awaiting response... 302 Found
Location: http://get.symfony.com/Symfony_Standard_Vendors_2.6.1.tgz [following]
--2014-12-24 01:31:31--  http://get.symfony.com/Symfony_Standard_Vendors_2.6.1.tgz
Resolving get.symfony.com (get.symfony.com)... 54.231.136.69
Connecting to get.symfony.com (get.symfony.com)|54.231.136.69|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 5132534 (4.9M) [application/x-tar]
[...]

PR #177, opened 2015-07-13
Added support for debuging the request performed by Guzzle

javiereguiluz:
This is useful to spot connection issues for bugs such as #92.

### DIFF
--- a/src/Symfony/Installer/DownloadCommand.php
+++ b/src/Symfony/Installer/DownloadCommand.php
@@ -338,7 +338,7 @@ abstract class DownloadCommand extends Command
             return;
         }
 
-        $errorMessage  = wordwrap($requirement->getTestMessage(), $lineSize - 3, PHP_EOL.'   ').PHP_EOL;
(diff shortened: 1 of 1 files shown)


src/Symfony/Installer/DownloadCommand.php
```diff
@@ -177,15 +177,20 @@ protected function checkProjectName()
...
         // check if the client must use a proxy server
         if (!empty($_SERVER['HTTP_PROXY']) || !empty($_SERVER['http_proxy'])) {
             $proxy = !empty($_SERVER['http_proxy']) ? $_SERVER['http_proxy'] : $_SERVER['HTTP_PROXY'];
-            $options['proxy'] = $proxy;
+            $defaults['proxy'] = $proxy;
         }
 
-        return new Client($options);
+        // enable request debugging if the command is in verbosity mode
+        if ($this->output->getVerbosity() >= OutputInterface::VERBOSITY_VERBOSE) {
```
stof: I suggest doing this only for the VERBOSITY_DEBUG level

PR #182, opened 2015-09-09
Update Guzzle to version 5 to avoid PHP 7 issues

javiereguiluz:
`guzzlehttp/progress-subscriber` has been removed because it's declared as `DEPRECATED` and it requires Guzzle 4.

### DIFF
--- a/src/Symfony/Installer/DownloadCommand.php
+++ b/src/Symfony/Installer/DownloadCommand.php
@@ -89,9 +89,12 @@ abstract class DownloadCommand extends Command
 
         /** @var ProgressBar|null $progressBar */
         $progressBar = null;
-        $downloadCallback = function ($size, $downloaded, $client, $request, Response $response) use (&$progressBar) {
-            // Don't initialize the progress bar for redirects as the size is much smaller
-            if ($response->getStatusCode() >= 300) {
+        $downloadCallback = function (ProgressEvent $event) use (&$progressBar) {
+            $downloadSize = $event->downloadSize;
+            $downloaded = $event->downloaded;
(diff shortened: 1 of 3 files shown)


src/Symfony/Installer/DownloadCommand.php
```diff
@@ -89,9 +88,12 @@ protected function download()
...
         /** @var ProgressBar|null $progressBar */
         $progressBar = null;
-        $downloadCallback = function ($size, $downloaded, $client, $request, Response $response) use (&$progressBar) {
-            // Don't initialize the progress bar for redirects as the size is much smaller
-            if ($response->getStatusCode() >= 300) {
+        $downloadCallback = function (ProgressEvent $event) use (&$progressBar) {
+            $size = $event->downloadSize;
+            $downloaded = $event->downloaded;
+
+            if ($size < 1 * 1024 * 1024) {
```
Pierstoval: Is the first `1` really necessary?

javiereguiluz: I guess not :smile: It's just a personal preference to read it more clearly as "1 megabyte". Should I change it?

Pierstoval: A small comment above the instruction might be more verbose about this :)

stof:
if it works fine, :+1:

stof:
Looks like Box cannot be installed on HHVM.

what could be done is installing and running Box with php 5.6 all the time (you can access it as ``~/.phpenv/versions/5.6/bin/php`` even when it is not the selected version)

Issue #161, opened 2015-05-18
[RFC] Integration Tests


javiereguiluz:
The problem
-----------

Recently we started recommending to use HTTPS for the installer. Sadly, the server infrastructure is not completely ready and this doesn't work. To make things worse, I didn't test the HTTPS feature properly before recommending it.

The solution
------------

The Symfony Installer is an essential tool for Symfony project and it must always work as expected. That's why, besides unit tests, we need **integration tests** to run them for each commit/pull request submitted to this repository.

Specifically, we need to do the following for any new code added to this project:

  1. Build the PHAR file with Box project.
  2. Install Symfony using HTTP and any of the available options (branch, specific version, `lts` special version, etc.)
  3. Install Symfony using HTTPS and any of the available options (branch, specific version, `lts` special version, etc.)
  4. Verify all the previous installations executing at least `php app/console --version` command in the installed Symfony application.

**How can we implement these integration tests?**

jakzal:
What would help in this situation is a smoke test. This could even be a simple bash script that builds the phar and verifies it can be executed with **few** variations. 

We should make it clear that only communication with the outside world should be tested here. This kind of tests are slow and we should have as little of them as possible.

desarrolla2:
I think that you can use symfony/proccess + phpunit to do this.

```php
<?php
use Symfony\Component\Process\Process;

class InstallerTest extends \PHPUnit_Framework_TestCase
{
    public function setUp()
    {
        // set permissions
    }

    public function tearDown()
    {
        // remove files
    }

    public function testBuildPhar()
    {
        $process = new Process('build-phar-file-here');
        $process->run();
        $this->assertTrue($process->isSuccessful());
        
        // verify checksum ? 
    }

    /**
     * @dataProvider branchesDataProvider
     *
     * @param string $branch
     */
    public function testNewProjectWithBranch($branch)
    {
        $process = new Process(sprintf('symfony new my_project %s', $branch));
        $process->run();
        $this->assertTrue($process->isSuccessful());

        // other verifications
    }
}
```

Pierstoval:
Need to ensure it doesn't create too many connections to the `symfony.com` website I guess

94noni:
Can we not get a master.zip of the installer from github website instead?
In a simple bash script: unzip it, install deps, build phar, and check difference commands for expected outputs (and version/branch/self-update)

stof:
@94noni I don't understand your comment

javiereguiluz:
[...]

PR #183, opened 2015-09-09
Added some integration tests

javiereguiluz:
In #161 we discussed a bit about integration tests. I know that some people are not very fond of this kind of tests, but the Symfony Installer is a critical piece of our platform and it must always work as expected.

This pull request adds a test which actually builds the PHAR file and test all the different installer options (no Symfony version, special `lts` version, branch version, etc.) It also tests the Symfony Demo application installation. No mocks or tricks are used, so the test actually downloads all the real Symfony packages from the Internet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+language: php
+sudo: false
+
+cache:
+  directories:
+    - $HOME/.composer/cache/files
+
+matrix:
+  fast_finish: true
+  include:
+    - php: 5.3
+    - php: 5.4
+    - php: 5.5
+    - php: 5.6
+    - php: hhvm
+    - php: nightly
+  allow_failures:
+    - php: hhvm
+
+install:
+  - composer install
+
+script:
+  - phpunit -c app
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,14 +8,12 @@ cache:
 matrix:
   fast_finish: true
   include:
-    - php: 5.3
     - php: 5.4
     - php: 5.5
     - php: 5.6
     - php: 7.0
     - php: hhvm
   allow_failures:
-    - php: 7.0
     - php: hhvm
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ matrix:
 
 before_install:
   - composer self-update
-  - curl -LSs http://box-project.github.io/box2/installer.php | php
+  - curl -LSs https://box-project.github.io/box2/installer.php | php
   - mv box.phar box
   - chmod 755 box
 
@@ -28,7 +28,3 @@ install:
 
 script:
   - phpunit
-
-after_script:
-  - rm box
-  - rm symfony.phar
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ matrix:
     - php: 7.0
     - php: hhvm
   allow_failures:
+    - php: 7.0
     - php: hhvm
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,9 +20,14 @@ before_install:
   - curl -LSs http://box-project.github.io/box2/installer.php | php
   - mv box.phar box
   - chmod 755 box
+  - php box build
 
 install:
   - composer install
 
 script:
   - phpunit
+
+after_script:
+  - rm box
+  - rm symfony.phar
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,13 +17,14 @@ matrix:
     - php: hhvm
 
 before_install:
+  - composer self-update
   - curl -LSs http://box-project.github.io/box2/installer.php | php
   - mv box.phar box
   - chmod 755 box
-  - php box build
 
 install:
   - composer install
+  - php box build
 
 script:
   - phpunit
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,18 @@ matrix:
     - php: 5.4
     - php: 5.5
     - php: 5.6
+    - php: 7.0
     - php: hhvm
-    - php: nightly
   allow_failures:
     - php: hhvm
+
+before_install:
+  - curl -LSs http://box-project.github.io/box2/installer.php | php
+  - mv box.phar box
+  - chmod 755 box
 
 install:
   - composer install
 
 script:
-  - phpunit -c app
+  - phpunit
--- a/composer.json
+++ b/composer.json
@@ -18,11 +18,11 @@
         "guzzlehttp/progress-subscriber": "~1.1",
         "symfony/console": "~2.5",
         "symfony/filesystem": "~2.5",
-        "raulfraile/distill": "~0.9,!=0.9.3,!=0.9.4",
-        "symfony/process": "^2.7"
+        "raulfraile/distill": "~0.9,!=0.9.3,!=0.9.4"
     },
     "require-dev": {
-        "kherge/box": "~2.5"
+        "kherge/box": "~2.5",
+        "symfony/process": "~2.5"
     },
     "extra": {
         "branch-alias": {
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,11 @@
         "guzzlehttp/progress-subscriber": "~1.1",
         "symfony/console": "~2.5",
         "symfony/filesystem": "~2.5",
-        "raulfraile/distill": "~0.9,!=0.9.3,!=0.9.4"
+        "raulfraile/distill": "~0.9,!=0.9.3,!=0.9.4",
+        "symfony/process": "^2.7"
+    },
+    "require-dev": {
+        "kherge/box": "~2.5"
     },
     "extra": {
         "branch-alias": {
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,6 @@
         "raulfraile/distill": "~0.9,!=0.9.3,!=0.9.4"
     },
     "require-dev": {
-        "kherge/box": "~2.5",
         "symfony/process": "~2.5"
     },
     "extra": {
--- a/src/Symfony/Installer/Tests/IntegrationTest.php
+++ b/src/Symfony/Installer/Tests/IntegrationTest.php
@@ -1,0 +1,118 @@
+<?php
+
+/*
+ * This file is part of the Symfony Installer package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Installer\Tests;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Process\Process;
+use Symfony\Component\Filesystem\Filesystem;
+
+class IntegrationTest extends \PHPUnit_Framework_TestCase
+{
+    private static $rootDir = __DIR__.'/../../../../';
+    private $fs;
+
+    public function setUp()
+    {
+        $this->fs = new Filesystem();
+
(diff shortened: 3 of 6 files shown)


composer.json
```diff
@@ -20,6 +20,10 @@
         "symfony/filesystem": "~2.5",
         "raulfraile/distill": "~0.9,!=0.9.3,!=0.9.4"
     },
+    "require-dev": {
+        "kherge/box": "~2.5",
```
stof: IMO, box should be installed separately (downloading their phar). Otherwise, you will get a bunch of unrelated deps in the phar (Box itself), and it makes things harder to test, as the phar should actually be built without dev deps (I hope the phar distributed for releases does not install dev dependencies in the phar)

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+
+matrix:
+  fast_finish: true
+  include:
+    - php: 5.3
+    - php: 5.4
+    - php: 5.5
+    - php: 5.6
+    - php: hhvm
+    - php: nightly
```
stof: using 7.0 would be better than nightly, as we will hopefully have 7.0 releases available on Travis in their next environment update (for now, it also installs the nightly) and nightly would become 7.1 once the PHP team creates a separate branch

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+  allow_failures:
+    - php: hhvm
+
+install:
+  - composer install
+
+script:
+  - phpunit -c app
```
stof: ``-c app`` is wrong. The config file is at the root

src/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -0,0 +1,118 @@
...
+/*
+ * This file is part of the Symfony Installer package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Installer\Tests;
```
stof: I suggest putting the tests in ``tests/`` at the root rather than in ``src/``. It will make it much easier to exclude the testsuite from the phar.
In Symfony itself, we have the tests in the codebase because we want to bring them in subtree splits, and so it was necessary to have them in the same subtree. And we have them in a subfolder of the code in bundles because of the convention for bundles being created in the pre-PSR-4 times. It is now possible to create repos for bundles with tests and source separated (Symfony does not actually care about it, as it will never look for the tests anyway)

src/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -0,0 +1,118 @@
...
+{
+    private static $rootDir = __DIR__.'/../../../../';
+    private $fs;
+
+    public function setUp()
+    {
+        $this->fs = new Filesystem();
+
+        $this->fs->remove(self::$rootDir.'/symfony.phar');
+        $this->runCommand('./vendor/bin/box build');
```
stof: I suggest building the phar only once using ``setUpBeforeClass`` instead of building it again for each test. It should make things faster

src/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -0,0 +1,118 @@
...
+    {
+        $projectDir = sprintf('%s/my_test_project', sys_get_temp_dir());
+        $this->fs->remove($projectDir);
+
+        $commonArguments = sprintf('new %s', $projectDir);
+        $output = $this->runCommand(sprintf('./symfony.phar %s %s', $commonArguments, $additionalArguments));
+        $this->assertContains('Downloading Symfony...', $output);
+        $this->assertRegExp($messageRegexp, $output);
+
+        $output = $this->runCommand(sprintf('cd %s && php app/console --version', $projectDir));
```
stof: you need to escape the project dir in case there is some spaces in it (or other special chars)

src/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * @dataProvider provideSymfonyInstallationData
+     */
+    public function testSymfonyInstallation($additionalArguments, $messageRegexp, $versionRegexp)
+    {
+        $projectDir = sprintf('%s/my_test_project', sys_get_temp_dir());
+        $this->fs->remove($projectDir);
+
+        $commonArguments = sprintf('new %s', $projectDir);
```
stof: the project dirshould be escaped here too

src/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -0,0 +1,118 @@
...
+        $output = $this->runCommand(sprintf('./symfony.phar %s %s', $commonArguments, $additionalArguments));
```
stof: you should use php explicitly instead of using the shebang, to make the test compatible with windows

src/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -0,0 +1,118 @@
...
+     * @throws RuntimeException in case the command execution is not successful
+     */
+    private function runCommand($command)
+    {
+        $process = new Process($command);
+        $process->setWorkingDirectory(self::$rootDir);
+        $process->run();
+
+        if (!$process->isSuccessful()) {
+            throw new \RuntimeException($process->getErrorOutput());
```
stof: use the ProcessFailedException, to give a better exception message on failure (you can even use ``$process->mustRun()`` rather than ``$process->run()``)

src/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Runs the given string as a command and returns the resulting output.
+     * The CWD is set to the root project directory to simplify command paths.
+     *
+     * @param string $command
+     *
+     * @return string
+     *
+     * @throws RuntimeException in case the command execution is not successful
```
stof: missing ``\``

src/Symfony/Installer/Tests/IntegrationTest.php
javiereguiluz: The problem with `setUpBeforeClass` is that it's static, so it causes us some troubles. The same happens with tearDownAfterClass() and that's why I needed to make `$rootDir` static, which is something I don't like at all.

src/Symfony/Installer/Tests/IntegrationTest.php
javiereguiluz: But in out case, `$projectDir` is always generated as `sys_get_temp_dir().'/my_test_project'`  Do you thin kit's necessary to escape it?

javiereguiluz:
@fabpot when you have some free time, please enable Travis for this repository. Thanks!

fabpot:
@javiereguiluz done

javiereguiluz:
Tests are finally green! @stof do you have any other comment before the merge? Thanks!

src/Symfony/Installer/Tests/IntegrationTest.php
stof: sys_get_temp_dir might be a folder with spaces in its path

src/Symfony/Installer/Tests/IntegrationTest.php
stof: well, the root dir would be a constant, if this was supported on 5.5+

.travis.yml
```diff
@@ -0,0 +1,34 @@
...
+install:
+  - composer install
+  - php box build
+
+script:
+  - phpunit
+
+after_script:
+  - rm box
+  - rm symfony.phar
```
stof: this is not needed. the Travis VM is deleted anyway

tests/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -0,0 +1,111 @@
...
+            throw new \RuntimeException(sprintf("Before running the tests, make sure that the Symfony Installer is available as a 'symfony.phar' file in the '%s' directory.", $this->rootDir));
+        }
+    }
+
+    public function testDemoApplicationInstallation()
+    {
+        $projectDir = sprintf('%s/my_test_project', sys_get_temp_dir());
+        $this->fs->remove($projectDir);
+
+        $output = $this->runCommand(sprintf('php symfony.phar demo %s', $projectDir));
```
stof: you still need to escape the project dir

tests/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -0,0 +1,111 @@
...
+        $this->assertContains('Downloading the Symfony Demo Application', $output);
+        $this->assertContains('Symfony Demo Application was successfully installed.', $output);
+
+        $output = $this->runCommand(sprintf('cd %s && php app/console --version', $projectDir));
```
stof: this needs escaping here too.

stof: And rather than using ``cd``, which not making the working directory configurable in your ``runCommand`` helper (defaulting to the root dir)

tests/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * @dataProvider provideSymfonyInstallationData
+     */
+    public function testSymfonyInstallation($additionalArguments, $messageRegexp, $versionRegexp)
+    {
+        $projectDir = sprintf('%s/my_test_project', sys_get_temp_dir());
+        $this->fs->remove($projectDir);
+
+        $commonArguments = sprintf('new %s', $projectDir);
```
stof: missing escaping of the project dir. And ``new`` could go in the command below instead of creating ``$commonArguments``

.travis.yml
```diff
@@ -0,0 +1,34 @@
...
+    - php: 5.5
+    - php: 5.6
+    - php: 7.0
+    - php: hhvm
+  allow_failures:
+    - php: hhvm
+
+before_install:
+  - composer self-update
+  - curl -LSs http://box-project.github.io/box2/installer.php | php
```
stof: you could use https here, no ?

javiereguiluz:
All comments have beeen addressed and tests are green. Merging!

tests/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -0,0 +1,111 @@
...
+
+namespace Symfony\Installer\Tests;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Process\Process;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+use Symfony\Component\Process\ProcessUtils;
+
+class IntegrationTest extends \PHPUnit_Framework_TestCase
```
stof: technically, this is a functional test :smile:

tests/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -0,0 +1,111 @@
...
+    private $rootDir;
+    private $fs;
+
+    public function setUp()
+    {
+        $this->rootDir = realpath(__DIR__.'/../../../../');
+        $this->fs = new Filesystem();
+
+        if (!$this->fs->exists($this->rootDir.'/symfony.phar')) {
+            throw new \RuntimeException(sprintf("Before running the tests, make sure that the Symfony Installer is available as a 'symfony.phar' file in the '%s' directory.", $this->rootDir));
```
stof: this should rather mark it as skipped, no ?

PR #159, opened 2015-05-11
Installation dev/beta versions (143)

94noni:
Hi

Here is my PR according to the comments on https://github.com/symfony/symfony-installer/pull/144 PR

Question:
- Is it possible for ```dev``` to be resolved like ```lts``` argument? (see https://github.com/symfony/symfony-installer/pull/105#issuecomment-75243491 and http://symfony.com/versions.json ```"dev":"2.7.0-BETA1"```)

capture d ecran 2015-05-11 a 11 52 48
capture d ecran 2015-05-11 a 11 26 20

*Docs if PR accepted*
*Screenshot when BETA2 was not released yet*

PR #185, opened 2015-09-11
Improve the installation of -dev and -BETA versions

javiereguiluz:
This finishes #159, which is related to #144 and #143.

The new behavior of the installer:

1) When installing a regular Symfony version, nothing changes from the previous installer.

2) When you install a `beta` version, you see a warning message:

symfony-unstable-warning

3) When you try to install a `dev` version, you get a better error message and a link to the article that will solve your problem:

symfony-dev-version

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -171,6 +171,11 @@ class NewCommand extends DownloadCommand
             ));
         }
 
+        // warn the user when downloading an unstable version
+        if (preg_match('/^.*\-(dev|BETA)\d*$/', $this->version)) {
+            $this->output->writeln("\n <bg=red> WARNING </> You are downloading an unstable Symfony version.");
+        }
+
         return $this;
     }
 
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -113,7 +113,7 @@ class NewCommand extends DownloadCommand
 
         // validate semver syntax
         if (!preg_match('/^2\.\d(?:\.\d{1,2})?(-(dev|BETA)\d*)?$/', $this->version)) {
-            throw new \RuntimeException('The Symfony version should be 2.N or 2.N.M, where N = 0..9 and M = 0..99.');
+            throw new \RuntimeException('The Symfony version should be 2.N or 2.N.M, where N = 0..9 and M = 0..99. The special "-dev" or "-BETA" suffixes are also supported.');
         }
 
         if (preg_match('/^2\.\d$/', $this->version)) {
@@ -171,8 +171,20 @@ class NewCommand extends DownloadCommand
             ));
         }
 
+        // "-dev" versions are not supported because Symfony doesn't provide packages for them
+        if (preg_match('/^.*\-dev$/', $this->version)) {
+            throw new \RuntimeException(sprintf(
+                "The selected version (%s) cannot be installed because it hasn't\n".
+                "been published as a package yet. Read the following article for\n".
+                "an alternative installation method:\n\n".
+                "> How to Install or Upgrade to the Latest, Unreleased Symfony Version\n".
+                "> http://symfony.com/doc/current/cookbook/install/unstable_versions.html",
+                $this->version
+            ));
+        }
+
         // warn the user when downloading an unstable version
-        if (preg_match('/^.*\-(dev|BETA)\d*$/', $this->version)) {
+        if (preg_match('/^.*\-BETA\d*$/', $this->version)) {
             $this->output->writeln("\n <bg=red> WARNING </> You are downloading an unstable Symfony version.");
         }
 
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -111,8 +111,8 @@ class NewCommand extends DownloadCommand
         }
 
         // validate semver syntax
-        if (!preg_match('/^2\.\d(?:\.\d{1,2})?(-(dev|BETA\d*))?$/i', $this->version)) {
-            throw new \RuntimeException('The Symfony version should be 2.N or 2.N.M, where N = 0..9 and M = 0..99. The special "-dev" or "-BETA" suffixes are also supported.');
+        if (!preg_match('/^2\.\d(?:\.\d{1,2})?(?:-(?:dev|BETA\d*|RC\d*))?$/i', $this->version)) {
+            throw new \RuntimeException('The Symfony version should be 2.N or 2.N.M, where N = 0..9 and M = 0..99. The special "-dev" "-BETA" and "-RC" suffixes are also supported.');
         }
 
         if (preg_match('/^2\.\d$/', $this->version)) {
@@ -183,8 +183,10 @@ class NewCommand extends DownloadCommand
         }
 
         // warn the user when downloading an unstable version
-        if (preg_match('/^.*\-BETA\d*$/i', $this->version)) {
+        if (preg_match('/^.*\-(BETA|RC)\d*$/i', $this->version)) {
             $this->output->writeln("\n <bg=red> WARNING </> You are downloading an unstable Symfony version.");
+            // versions provided by the download server are case sensitive
+            $this->version = strtoupper($this->version);
         }
 
         return $this;
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -111,7 +111,7 @@ class NewCommand extends DownloadCommand
         }
 
         // validate semver syntax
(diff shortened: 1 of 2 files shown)


javiereguiluz:
I'd like to merge this change soon because I want to test it well before we enter into the most intense release period in Symfony history (including a major version change). Please add any comment or review that you may have. Thanks!

src/Symfony/Installer/NewCommand.php
```diff
@@ -105,19 +105,14 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         // 'lts' is a special version name that refers to the current long term support version
-        if ('lts' === $this->version) {
+        if (in_array($this->version, array('latest', 'lts'))) {
             return $this;
         }
 
         // validate semver syntax
-        if (!preg_match('/^2\.\d(?:\.\d{1,2})?$/', $this->version)) {
-            throw new \RuntimeException('The Symfony version should be 2.N or 2.N.M, where N = 0..9 and M = 0..99');
+        if (!preg_match('/^2\.\d(?:\.\d{1,2})?(-(dev|BETA)\d*)?$/', $this->version)) {
```
xabbuh: Numbers don't make sense after `dev`, do they? And what about lowercased `beta` strings?

xabbuh:
@javiereguiluz SO we can now download archives of all beta versions? Does this include all beta versions that have been released in the past?

javiereguiluz:
@xabbuh yes, we can download any past beta package. For example: `symfony new my_project 2.6.0-BETA1`

src/Symfony/Installer/NewCommand.php
javiereguiluz: In 9b3cdc1 we no longer allow numbers after the `dev` version.

Regarding the `BETA` uppercase/lowercase issue, I thought about that but I decided to leave it untouched. If we silently fix `beta` as `BETA`, we should also fix `DEV` to `dev`. Honestly, I don't know if this is right for user experience or it's bad because we are supporting wrong version names.

weaverryan:
:+1: as is - it'll be great to be able to show people how to create a new beta project for testing when the beta's are released.

src/Symfony/Installer/NewCommand.php
xabbuh: Composer actually makes no difference here. So we would be consistent with its behaviour.

javiereguiluz:
After @xabbuh's comments, I can confirm that Composer is case insensitive at least on Mac OS. These commands result in the same installation:

```
$ composer create-project symfony/framework-standard-edition my_project 2.7.0-BETA1
$ composer create-project symfony/framework-standard-edition my_project 2.7.0-beta1
```

That's why in 31765ca I've changed all `dev` and `BETA` comparisons to case insentivity. I'll merge this last version unless someone tells me not to. Thanks!

stof:
> I can confirm that Composer is case insensitive at least on Mac OS

it is on all systems. It normalizes the case of stability specifiers: https://github.com/composer/composer/blob/15face5432d7b7334db6ac69fac0190971cafa6e/src/Composer/Package/Version/VersionParser.php#L49

src/Symfony/Installer/NewCommand.php
```diff
@@ -105,19 +105,14 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         // 'lts' is a special version name that refers to the current long term support version
-        if ('lts' === $this->version) {
+        if (in_array($this->version, array('latest', 'lts'))) {
             return $this;
         }
 
         // validate semver syntax
-        if (!preg_match('/^2\.\d(?:\.\d{1,2})?$/', $this->version)) {
-            throw new \RuntimeException('The Symfony version should be 2.N or 2.N.M, where N = 0..9 and M = 0..99');
+        if (!preg_match('/^2\.\d(?:\.\d{1,2})?(-(dev|BETA\d*))?$/i', $this->version)) {
```
stof: you should use non-capturing groups here

stof:
@javiereguiluz what about ``RC`` versions ?

stof:
@javiereguiluz shouldn't you normalize the case of ``BETA`` for the generated archive URL though ? Or is it case insensitive on the server too ?

src/Symfony/Installer/NewCommand.php
javiereguiluz: I've changed it by this: `'/^2\.\d(?:\.\d{1,2})?(?:-(?:dev|BETA\d*))?$/i'`

javiereguiluz:
In https://github.com/symfony/symfony-installer/commit/b26b217c24c6ee713694e633391c2ac5f6af24ca I've added support for RC versions and I've normalized the version names before downloading them from the server.

src/Symfony/Installer/NewCommand.php
```diff
@@ -105,19 +105,14 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         // 'lts' is a special version name that refers to the current long term support version
-        if ('lts' === $this->version) {
+        if (in_array($this->version, array('latest', 'lts'))) {
             return $this;
         }
 
         // validate semver syntax
-        if (!preg_match('/^2\.\d(?:\.\d{1,2})?$/', $this->version)) {
-            throw new \RuntimeException('The Symfony version should be 2.N or 2.N.M, where N = 0..9 and M = 0..99');
+        if (!preg_match('/^2\.\d(?:\.\d{1,2})?(?:-(?:dev|BETA\d*|RC\d*))?$/i', $this->version)) {
```
xabbuh: You could now pass a patch version and a dev suffix. I think both parts should be exclusive.

stof: dev suffixes are all rejected just later, so this is not a big deal IMO

xabbuh: Yeah, but this does not only apply to the `-dev` suffix but also to `-BETA` and `-RC`.

stof: and beta versions are named 2.1.0-beta, not 2.1-beta

xabbuh: Hm indeed, then we can move that part of the regex.

javiereguiluz: I've changed this regular expression by:

```php
if (!preg_match('/^2\.\d(?:\.\d{1,2})?|2\.\d(?:-(?:dev|BETA\d*|RC\d*))$/i', $this->version))
```

Please review it. Thanks!

xabbuh: Did you forget to push the changes?

tests/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -106,6 +106,12 @@ public function provideSymfonyInstallationData()
                 '/.*Symfony 2\.5\.6 was successfully installed.*/',
                 '/Symfony version 2\.5\.6 - app\/dev\/debug/',
             ),
+
+            array(
+                '2.7-BETA1',
```
xabbuh: This version should fail while `2.7.0-BETA1` should succeed (I think it would be good to have tests for both versions).

src/Symfony/Installer/NewCommand.php
javiereguiluz: I've pushed it .... and then reverted it. The correct beta version is `x.y.z-BETAn`, not `x.y-BETAn`

tests/Symfony/Installer/Tests/IntegrationTest.php
javiereguiluz: Indeed. Fixed. Thanks.

tests/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -106,6 +106,12 @@ public function provideSymfonyInstallationData()
                 '/.*Symfony 2\.5\.6 was successfully installed.*/',
                 '/Symfony version 2\.5\.6 - app\/dev\/debug/',
             ),
+
+            array(
+                '2.7.0-BETA1',
+                '/.*Symfony 2\.7\.0\-BETA1 was successfully installed.*/',
+                '/Symfony version 2\.7\-BETA1 - app\/dev\/debug/',
```
xabbuh: This seems to need an update too, doesn't it?

javiereguiluz: I'm trying to do things fast ... but I do everything wrong :(  Thanks for reviweing this!

src/Symfony/Installer/NewCommand.php
xabbuh: There would still be some versions that pass the regex but are not valid. Though actually the installer will fail with a meaningful message anyway later on. So I think it's okay if we just stick with this solution.

xabbuh:
:+1:

javiereguiluz:
@xabbuh thanks for the review. I'm gonna merge it now, so we have time to test it well before the 2.8.0-BETA1 release. Thanks!

PR #196, opened 2015-11-13
Updated the hash generator for the new content-hash Composer property

### LINKED ISSUE #195: The new command needs to update the content-hash in the lock file
The composer.lock now contains 2 hashes:

- the old ``hash`` property which hashes the full composer.json file
- the new ``content-hash`` property, which hashes only the settings relevant to the resolution algorithm

New versions of composer will ignore the hash if the content-hash is there, avoiding to show a warning in case you change the composer.json in a non-meaningful way (changing the formatting, changing the suggest section, etc...).

As the installer updates the name in the composer.json and this name is taken into account by the content-hash, it needs to update the content-hash too, not only the hash.

javiereguiluz:
This fixes #195.

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -440,32 +440,32 @@ class NewCommand extends DownloadCommand
      */
     private function syncComposerLockFile()
     {
-        $composerFileContents = file_get_contents($this->projectDir.'/composer.json');
-        $lockFileContents = json_decode(file_get_contents($this->projectDir.'/composer.lock'), true);
+        $composerJsonFileContents = file_get_contents($this->projectDir.'/composer.json');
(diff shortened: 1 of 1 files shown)


src/Symfony/Installer/NewCommand.php
```diff
@@ -443,8 +443,56 @@ private function syncComposerLockFile()
...
+            $lockFileContents['content-hash'] = $this->getComposerContentHash($composerFileContents);
+        }
 
         file_put_contents($this->projectDir.'/composer.lock', json_encode($lockFileContents, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES)."\n");
     }
+
+    /**
+     * Returns the md5 hash of the sorted content of the composer file.
+     *
+     * @see  https://github.com/composer/composer/blob/master/src/Composer/Package/Locker.php#L394-L431
```
xabbuh: This URL looks wrong.

Issue #190, opened 2015-10-24
composer.json is Empty after instalation of Symfony


sinner:
composer.json file is empty after instalation of Symfony Standar Edition 2.7.

I run on the terminal:

```sh
symfony new proyectName
```

The version I used was 1.1.7

OS: Windows 10

Solution: Change to Linux :+1: I switched to Ubuntu

WouterJ:
This bug doesn't occur on Windows 8.1. I'll debug it later today on a Windows 10 PC.

WouterJ:
I can't reproduce this on Windows 10. Everything works like it should.

xabbuh:
@WouterJ Does the behaviour change when you require `2.7.5` explicitly (maybe the issue is resolved with the build of the new release and only the archive for the version before was broken).

senenh:
@WouterJ  Is there any way I can help you seeing that it does not create the composer properly for me?

xabbuh:
@senenh Can you paste the complete output of the installer?

senenh:
@xabbuh 

Do you men this?:

c:\Apache24\htdocs>php symfony new test

 Downloading Symfony...

 Preparing project...

 OK  Symfony 2.7.6 was successfully installed. Now you can:

    * Change your current directory to C:\Apache24\htdocs\test

    * Configure your application in app/config/parameters.yml file.

    * Run your application:
        1. Execute the php app/console server:run command.
        2. Browse to the http://localhost:8000 URL.

    * Read the documentation at http://symfony.com/doc

WouterJ:
Hmm, the archives on symfony.com contain the proper composer.json files, so that's not the problem.

@sinner can you maybe check this project out locally (it's easier to debug):
```
C:\Apache24\htdocs> git clone http://github.com/symfony/symfony-installer
```
And then add some debug lines in thsi method: https://github.com/symfony/symfony-installer/blob/master/src/Symfony/Installer/NewCommand.php#L327-L361 (espescially the contents of `$contents` and the `json_encode` function injected into `file_put_contents`). And then execute the locally checked out installer:
```
C:\Apache24\htdocs> php symfony-installer/symfony new test
```

I wonder what's making the file empty and, unless I'm missing something, that's the only method editing this file.

senenh:
Now I get the following. Maybe is something easy but I'm little lost with this error.

c:\Apache24\htdocs>php symfony-installer/symfony new test1

 Downloading Symfony...




[...]

PR #197, opened 2015-11-16
Fix lowercase issue with unicode characters

Pierstoval:
As of #190, @cccm62 pointed the fact that using `symfony new` with a name containing unicode characters, like `symfony new génial`, the package name were converted to `gã©nial` due to the lack of utf-8 support in `strtolower`. 
 
This is why I added the `u` flag to `preg_replace` to support unicode, added the `mb_strtolower` function if it exists, and added a workaround if not.
 
Does it sound correct?

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -397,8 +397,8 @@ class NewCommand extends DownloadCommand
     {
         return strtolower(
             preg_replace(
-                array('/([A-Z]+)([A-Z][a-z])/', '/([a-z\d])([A-Z])/'),
-                array('\\1-\\2', '\\1-\\2'),
+                array('~[^\w.-]~iu', '~([A-Z]+)([A-Z][a-z])~u', '~([a-z\d])([A-Z])~u'),
+                array('', '\\1-\\2', '\\1-\\2'),
                 strtr($name, '-', '.')
             )
         );


src/Symfony/Installer/NewCommand.php
```diff
@@ -415,13 +415,19 @@ protected function generateComposerProjectName()
...
+        $name = preg_replace(
+            array('/([A-Z]+)([A-Z][a-z])/u', '/([a-z\d])([A-Z])/u'),
+            array('\\1-\\2', '\\1-\\2'),
+            strtr($name, '-', '.')
         );
+
+        if (function_exists('mb_strtolower')) {
+            $name = mb_strtolower($name);
+        } else {
+            $name = utf8_encode(strtolower(utf8_decode($name)));
```
nicolas-grekas: this is not safe, it works only for the iso8859-1 subset of unicode.

Pierstoval: Actually, this is typically the kind of workaround I'm not sure about :confused: Do you have a better idea?

cccm62: The problem may be the name of the project, but especially the username added by `generateComposerProjectName` (in my case $_SERVER['USERNAME'])

And I tried, this change has not resolved the error.

Otherwise, I don't think using the username for the vendor is really necessary ... a quick solution would be to use only the last condition:

    } else {
        // package names must be in the format foo/bar
        $name = $name.'/'.$name;
    }

... or to replace all special chars (I don't think tha composer package name should contain accented chars)

Pierstoval: Replacing special chars sounds a much better idea, we'd just need to "slugify" the package name and remove any non-usable character.
Any idea on how to do this with no "heavy" dependency?

stof: accented chars are not valid in composer package names anyway: https://github.com/composer/composer/blob/d7d91269ec4fb0da171a1c38fa1a47c493627dc0/src/Composer/Package/Loader/ValidatingArrayLoader.php#L52

cccm62: return strtolower(preg_replace(
        array('/[^\da-z]/i', '/([A-Z]+)([A-Z][a-z])/', '/([a-z\d])([A-Z])/'),
        array('', '\\1-\\2', '\\1-\\2'),
        strtr($name, '-', '.')
    ));

cccm62: return strtolower(preg_replace(
        array('/[^\/\da-z]/i', '/([A-Z]+)([A-Z][a-z])/', '/([a-z\d])([A-Z])/'),
        array('', '\\1-\\2', '\\1-\\2'),
        strtr($name, '-', '.')
    ));

Pierstoval: @cccm62 I added your proposal, optimized it a little, and removed the unicode functions.
It now removes all non-word characters.

cccm62: Just another thought... don't know how USERNAME can look for non latin countries (i.e. Kanji ?), perhaps it would be safer to replace chars by underscores?

Or fix vendor/name separately (and ensure it is not empty)?

    protected function generateComposerProjectName()
    {
        $name = $this->projectName;
        
        if (!empty($_SERVER['USERNAME'])) {
            $vendor = $_SERVER['USERNAME'];
        } elseif (true === extension_loaded('posix') && $user = posix_getpwuid(posix_getuid())) {
            $vendor = $user['name'];
        } elseif (get_current_user()) {
            $vendor = get_current_user();
        } else {
            $vendor = $name;
        }
        $vendor = $this->fixComposerPackageName($name);
        $name = $this->fixComposerPackageName($name);
        
        // package names must be in the format foo/bar
        return $vendor.'/'.$name;
    }

javiereguiluz:
@nicolas-grekas sorry to bother you but ... do the latest changes of this Pull Request address the concerns that you expressed in your comment above? Thanks!

nicolas-grekas:
I don't know if it's correct (no replace will happen if the string is not a unicode string)
A few test cases might help deciding...

Pierstoval:
Hmm, sounds hard to create tests that are based on a server-side user name :confused: 
Feel free to fork this branch if you know how to do it, because I don't feel confident enough in working in more "low-level" code :confused: .
 
Or, additionnally, we can try to "detect" any inconsistency in the `username` and then propose a special `--package-user-name` or similar option that will be used as the package owner name in the composer.json file, what do you think?

javiereguiluz:
Closing in favor of a new proposal in #278.

Issue #124, opened 2015-03-29
Add proxy support


javiereguiluz:
Some people are asking us (via Twitter, email and symfony blog comments) to add support for using the installer behind a company proxy. I think it's a reasonable request and we should provide this feature if it's not too complicated.

Apparently, Composer provides proxy support just by [relying on some environment variables](https://getcomposer.org/doc/03-cli.md#http-proxy-or-http-proxy). Since we use a modern Guzzle version, where [proxy support is built-in and trivial to use](http://guzzle.readthedocs.org/en/latest/clients.html#proxy), do you think that would be enough to provide proxy support? Does this also apply to Windows users?

raulfraile:
@javiereguiluz :+1: I also think that is an important feature as many people may need it. Even though the `env` variable would be enough, I suggest adding an optional argument to the command to make it easier to use, override the value of the `env` variable or create aliases. The command should also inform clearly that is using a proxy.

blag001:
:+1: 
Need this Feature, for now I can't do anything of the new installer without the proxy support :/
For linux users, the classic way is to work with `http_proxy= | https_proxy= | ftp_proxy=` env variable.
(& for ubuntu user you can look for a network-manager proxy config)
But a `--proxy ` param like CURL will be fine too...
Thanks !

javiereguiluz:
I'm going to work on this and I'l submit a PR soon.

javiereguiluz:
There is a pull request (#132) ready to fix this issue. It would be great if real proxy server users could test it before merging it. To do so, just execute these commands in your system:

```bash
$ mkdir temp
$ cd temp/
$ git clone https://github.com/symfony/symfony-installer.git
$ cd symfony-installer/
$ git checkout -b test_proxy
$ git pull https://github.com/javiereguiluz/symfony-installer.git fix_124

$ ./symfony new my_project
```

raulfraile:
@javiereguiluz it works fine using an anonymous proxy (I added a `var_dump()` to make sure it was setting the `proxy` setting in Guzzle):

```
$ export HTTP_PROXY=181.88.177.145:8080
$ php symfony new my_project9
```

![screen shot 2015-04-01 at 16 06 01](https://cloud.githubusercontent.com/assets/595160/6942627/230c20e0-d889-11e4-9cc9-19746442e97a.png)

It is much more slower, but it's normal as we are using a proxy.

More anonymous proxies: http://proxylist.hidemyass.com/

[...]

PR #200, opened 2015-11-21
Add cURL extension requirement

javiereguiluz:
According to these comments, this change can prevent some errors: https://github.com/symfony/symfony-installer/issues/124#issuecomment-128694119

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,7 @@
     },
     "require": {
         "php": ">=5.4.0",
+        "ext-curl": "*",
         "guzzlehttp/guzzle": "~5.0",
         "symfony/console": "~2.6",
         "symfony/filesystem": "~2.5",


weaverryan:
This certainly makes sense. But does this also cover the use-case of me downloading the installer phar file and trying to run it without curl? Or is that already handled.

ogizanagi:
Is curl really needed anyway ? (I mean [AFAIK, guzzle does not rely only on curl but as many different handlers available](http://docs.guzzlephp.org/en/latest/faq.html))
Maybe, instead of requiring curl, we could add a check through [this function or similar](https://github.com/guzzle/guzzle/blob/5c12b74e9405c50fb0fd6ab831dd6e9df5560b1f/src/functions.php#L104-L125) in order to ensure one of the needed handler exist in order to make the request, or throw an appropriated exception in case it isn't.

xabbuh:
Is the cURL extension available by default on Windows?

emil-nasso:
I tested this. cURL is not required.

I ran everything using docker with this dockerfile:

```
FROM debian:jessie
VOLUME .:/root/shared
WORKDIR /root/shared
RUN apt-get update && apt-get install -y php5-cli

ENTRYPOINT /bin/bash
```

i then built and ran it:

```
docker build -t symfony-test .
docker run -it -v `pwd`:/root/shared symfony-test /bin/bash
```

Out-of the box, that worked. I double-checked to make sure that curl was not installed.

When I set `allow_url_fopen = Off` in php.ini i got this error instead, when i ran the installer:



So (as it is implemented right now) guzzle in the symfony installer tries curl first and then fopen (as can be seen here: https://github.com/guzzle/guzzle/blob/5.3/src/Utils.php#L181 ).

So, to summarize: Curl is not required and you will get an error message if you lack curl and the `allow_url_open`-flag.

I wrote this (to replace this line: https://github.com/symfony/symfony-installer/blob/master/src/Symfony/Installer/DownloadCommand.php#L224 ) if we want to make the error message more "symfony" and a bit less confusing for new users:

```
        try {
            $handler = \GuzzleHttp\Utils::getDefaultHandler();
        } catch (\RuntimeException $e) {
            throw new \RuntimeException(
                'The symfony installer requires the php-curl extension or the '
                . 'allow_url_fopen ini setting.');
        }
        return new Client(array('defaults' => $defaults, 'handler' => $handler));
```

Should i create a pull-request or is the old message "good enough"?

javiereguiluz:
@emil-nasso thanks a lot for working on this issue and for your insightful explanation. I'm closing this issue as "won't fix". Besides, I like your proposal to display a custom error message. Please, consider submitting it as a pull request. Thanks!

PR #205, opened 2015-11-30
Moved the version in DownloadCommand

### LINKED ISSUE #204: Issue when installing the Symfony Demo with the latest Installer version
I'm getting this error but I don't have time at the moment to investigate it. Let's see if someone can take a look at this:

symfony_installer_error

Pierstoval:
This fixes #204

### DIFF
--- a/src/Symfony/Installer/DownloadCommand.php
+++ b/src/Symfony/Installer/DownloadCommand.php
@@ -36,13 +36,40 @@ use Symfony\Installer\Exception\AbortException;
  */
 abstract class DownloadCommand extends Command
 {
-    /** @var Filesystem */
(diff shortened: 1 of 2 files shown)


ogizanagi:
:+1: Always wondered why it was like it. BTW, you could also move `$downloadedFilePath` and `$requirementsErrors` IMO.

Also, it makes no sense the `DownloadCommand` has methods with `@return NewCommand` in the PhpDoc block and deals bad with autocompletion.

Pierstoval:
@ogizanagi done :)

javiereguiluz:
@Pierstoval thanks for providing such a fast bugfix!

PR #208, opened 2015-11-30
Fix symfony demo uses lts

ogizanagi:
- [x] Fix `symfony demo` command
> The demo application still uses LTS (2.7 actually), so the `symfony demo` command also fails ([search for `var/SymfonyRequirements.php`](https://travis-ci.org/symfony/symfony-installer/jobs/94038672#L209)). :/
It's solved by using `lts` as the default version for the `DemoCommand`.

- [x] Fix tests
> This PR also updates tests failures due to the new LTS and latest releases, in order to have everything green again.


javiereguiluz:
@ogizanagi thanks for this bugfix.

Issue #217, opened 2015-12-04
.gitignore installed by installer breaks application when shared


emil-nasso:
When installing symfony using `symfony new teststuff` it doesn't seem to use the .gitignore-file that can be found here: https://github.com/symfony/symfony-standard/blob/master/.gitignore

Instead it puts this into the file:

```
/app/config/parameters.yml
/bin/
/build/
/composer.phar
/vendor/
/web/bundles/
/var/
!var/cache/.gitkeep
!var/logs/.gitkeep
!var/sessions/.gitkeep
/phpunit.xml
```

This will break sharing of the application because the `bin` directory is ignored.

To reproduce:

* `symfony new test`
* `cd test; git init; git push {somewhere}`
* `cd /tmp; git clone {somewhere}; cd test`
* `composer install`

Composer will exit with errors because it can't find `app/console` or `bin/console`. It can be fixed by replacing the `.gitignore` with the one from symfony standard edition and commit/push the files from the original repository.

ogizanagi:
I agree the fact `.gitignore` are not in sync is an issue. That's why I opened #211

emil-nasso:
@ogizanagi yeah. I agree. And in this case, it actually breaks the application.

baiolo:
Behaviour confirmed.
Temporary workaround:

- Remove the installer .gitignore and replace with the one form [https://github.com/symfony/symfony-standard/blob/master/.gitignore](https://github.com/symfony/symfony-standard/blob/master/.gitignore).

- Copy the `bin` folder from a freshly installed Symfony project into the shared project.

ogizanagi:
This will be fixed once #218 is merged, and a new version of the installer released.

In the meanwhile, you can test this patch by building the phar yourself, or download it from this fake release on Github: https://github.com/ogizanagi/symfony-installer/releases/tag/synced_gitignore

stof:
Closing as #218 has been merged, and released in 1.5.0

Issue #211, opened 2015-12-01
.gitignore sync


ogizanagi:
I wonder why the installer have to recreate the `.gitignore` file instead of using the one provided in the `symfony/standard-edition` ?

The [build script](https://github.com/sensiolabs/SensioDistributionBundle/blob/master/Resources/bin/build.sh#L51) used to create the archive remove the `.gitignore` file. But it'll certainly be easier to maintain if was kept if the line was slightly different:
```diff
- rm -rf app/cache/* app/logs/* .git*
+ rm -rf app/cache/* app/logs/* .git/
```
_(Also I'm not sure `cache` and `logs` directories should still be removed ?)_

[This method](https://github.com/symfony/symfony-installer/blob/master/src/Symfony/Installer/DownloadCommand.php#L310-L353) could then be removed. Currently, the generated `.gitignore` file for 3.0 for instance is not in sync with the latest changes:

```diff
1a2
> /bin/
3,6c4,7
< /phpunit.xml
< /var/*
< !/var/cache
< /var/cache/*
---
> /composer.phar
> /vendor/
> /web/bundles/
> /var/
8,9d8
< !/var/logs
< /var/logs/*
11,12d9
< !/var/sessions
< /var/sessions/*
14,16c11
< !var/SymfonyRequirements.php
< /vendor/
< /web/bundles/
---
> /phpunit.xml
```

Pierstoval:
:+1: for this, the gitignore is already provided by the builder.
 
But it needs a new Symfony release for this :D

ogizanagi:
@Pierstoval : Not necessarily. Can't it be a new installer release only ? (And recreating an archive containing the `.gitignore` file for each maintained Symfony versions)

Pierstoval:
It would need a Symfony Standard release because the build script generates the tarball which contains the "built" datas, so the `.gitignore` file is not present now, but would be if the script is updated and a new version is released (because tarballs are pushed to `get.symfony.com` on each release).
It's not the installer that removes the `.gitignore` file :)
 
I don't know if the Symfony team can (or would) execute manually the tarball builder script, but actually I wonder if we could find a workaround in the installer, for instance download the `.gitignore` from github raw files in the `DownloadCommand`...

What do you think?

ogizanagi:
Personally, I feel like using a workaround in the installer would be useless, whereas updating the build script and the archives (and I mean recreate them through the updated script, of course) would be way more reliable.

[...]

PR #218, opened 2015-12-04
Keep .gitignore files in sync

ogizanagi:
This is only a quick fix regarding the issue mentioned in #211 , and the few recent issues related:

- #217
- https://github.com/symfony/symfony/issues/16815

### DIFF
--- a/src/Symfony/Installer/DownloadCommand.php
+++ b/src/Symfony/Installer/DownloadCommand.php
@@ -306,48 +306,46 @@ abstract class DownloadCommand extends Command
     }
 
     /**
-     * Creates the appropriate .gitignore file for a Symfony project.
+     * Creates the appropriate .gitignore file for a Symfony project if it doesn't exist.
      *
      * @return $this
      */
     protected function createGitIgnore()
     {
-        $gitIgnoreEntries = array(
-            '/app/config/parameters.yml',
-            '/bin/',
-            '/build/',
-            '/composer.phar',
-            '/vendor/',
-            '/web/bundles/',
-        );
-        if ($this->isSymfony3()) {
-            $gitIgnoreEntries = array_merge($gitIgnoreEntries, array(
-                '/var/',
-                '!var/cache/.gitkeep',
-                '!var/logs/.gitkeep',
-                '!var/sessions/.gitkeep',
-                '/phpunit.xml',
-            ));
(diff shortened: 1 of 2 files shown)


Pierstoval:
So you just wrote the "raw" .gitignore files that are used in both 2.x and 3.x branches then, it sounds better :+1:

ogizanagi:
@Pierstoval : Yes, and your comment also made me think the array syntax is useless now, so here is the nowdoc alternative.
Even if nothing is done about #211, it is then more straightforward to update.

Pierstoval:
I love this string format even if it looks a bit ugly in the first sight :D I'm :+1: on this

javiereguiluz:
I'm not sure about the new format. I think it's a bit ugly :(

ogizanagi:
@javiereguiluz: I can make those two ugly strings constants or dedicated files tonight if you think it's better. But I think the string format should be kept over the array entries for maintainability.

ogizanagi:
@javiereguiluz : PR updated with two new proposals:

- 091505d avoid ugly strings by adding both `.gitignore` files as resources in the phar archive (cannot let the leading dot, because box does not allow easily to not ignore dot files currently).
- 809cd1b retrieve `.gitinore` files directly from github. (If no connection available, it should generate an empty `.gitignore` file)  
IMHO, that's not the best option for the future, because I'd like to avoid extra HTTP requests, in case we add cache later, and allow to use the installer without any connection.  
So it doesn't fully solved #211, and I still think those files should be part of the generated archive.

javiereguiluz:
@ogizanagi thanks for working on this. I like the last proposal that retrieves the `.gitignore` from GitHub.

In order to solve the original error, I'm trying to modify the build script to maintain the `.gitignore` file (see https://github.com/sensiolabs/SensioDistributionBundle/pull/246). If this change gets approved, we'll have the `.gitignore` for **new** packages only, because the old packages won't be released.

That's why I ask you: should we add a check in your code to verify if the `.gitignore` file exists before trying to download it from GitHub? Thanks!

weaverryan:
:+1: for your idea Javier. I was wondering (until you explained it) why we weren't just using the `.gitignore` from the SE.

ogizanagi:
@javiereguiluz: Done.

Issue #220, opened 2015-12-05
Add a warning about old symfony installer version


jzawadzki:
Hi,

I noticed that on current symfony upgrade many users are rasing issues here that symfony installer doesn't work correctly. Most of those issues were already fixed but users haven't upgraded installer itself.

Do you think we should add a warning like composer have?
```
 Warning: This development build of composer is over 60 days old. It is recommended to update it by running "composer self-update" to get the latest version
```
 
And if so - how we should approach it? 
Composer compiles phar itself and adds constant with timestamp. We are just updating whole file (but only if there is newest version). 
Maybe we can just compare last modified time of phar with current time (+ `touch` file on `selfupdate` when its still up-to-date?)

Pierstoval:
There is already a discussion about this in #136 , and even if the issue is closed, actually one of @xabbuh's comment is still pending for response :)

ogizanagi:
I particulalry agree with @xabbuh's comment (https://github.com/symfony/symfony-installer/issues/136#issuecomment-89250313). I was not aware of this thread, and though about a similar idea with an error handler suggesting the user to update its installer if the version mismatches the latest one.

IMO, we've seen enough in the last days the need for such a feature. The installer should be pretty stable most of the time, but we cannot perfectly imagine how it would be in the future, or prevent all issues.

Pierstoval:
The best solution IMO is to ask SensioLabs to add a simple "installer_version" field in versions.json that would be updated on each release with a simple hook. It's possible, but it needs some setup :)

javiereguiluz:
The installer version is published in `get.symfony.com/symfony.version`, which is used by the Installer itself.

Pierstoval:
Ok, then it should be easy to do!

jzawadzki:
first draft up - please comment - and Happy SymfonyCon Hackday everyone :)

chepeftw:
Hello! I was working on an update for this issue too, I got problems with my wifi on Saturday so I was not able to commit, I was thinking in just adding a function to DownloadCommand class, and the add the call to that function in the chain of methods in the execute method for DemoCommand and NewCommand. Do you think is worth a 3rd PR for the same issue but different implementation?

Pierstoval:
I think we should first wait for @javiereguiluz to merge #221 before :)

xabbuh:
closing as this was fixed by #221

Issue #136, opened 2015-04-03
Add a warning message when the installer is not up-to-date.


Pierstoval:
I was thinking about the releases process in the last issues, and I think that actually, the users will not often update their installer "knowingly".
 
I wanted to introduce a new proposal: an information/warning message when the installer is not up-to-date, similar to what Composer does:
![capture d ecran 2015-04-03 a 11 49 55](https://cloud.githubusercontent.com/assets/3369266/6980776/96264ffa-d9f7-11e4-8f0e-d624d03bb880.png)
 
But here we are facing one problem:
Once the installer will have maaaaany features, maybe there will not be new release for weeks/months.

Here is the tags list and their release date

```yml
2014-11-05: v0.0.1
2014-11-19: v0.0.2 # 14 days later
2014-11-20: v0.0.5 # 1 day later
2014-11-21: v0.1.0 # 1 day later
2014-11-26: v0.2.0 # 5 days later
2014-12-04: v0.3.0 # 8 days later
2014-12-08: v0.4.0 # 4 days later
2014-12-16: v0.5.0 # 8 days later
2015-02-27: v1.0.0 # 73 days later (more than 2 months)
2015-03-26: v1.0.1 # 27 days later
2015-04-03: v1.1.0 # 8 days later
```
 
As they're no regular update, we can't rely on the same Composer's behavior with the 30-days time comparison.
 
I was thinking on a solution by calling the Github API to check for all tags, and check a hard-coded version right inside the phar file, in any command, to compare directly the user's version with the latest release available, and then throw a warning whenever they're different, or when the user's one is inferior to the remote one.
Edit: FYI, we can see all tags on the API right here : https://api.github.com/repos/symfony/symfony-installer/tags .
 
Then it would allow the user to be warned whenever there's a new update, and possibly invoke some curiosity to him for he can go on Github and see the release note.
 
What do you think?

javiereguiluz:
I thought about this feature and I discarded it for the same reasons that you pointed. In fact, the installer is "finished". We only fix reported bugs or add new features that are "mandatory" (such as the recent proxy server support).

In the future we'll only add features such as installing Symfony distributions (although it's not sure), caching of downloaded files (first we'll see if it's really needed) and file integrity verification through digital signatures (I'd like to add it but it looks like extremely complex to do it).

xabbuh:
[...]

PR #221, opened 2015-12-05
Checking that symfony installer is up-to-date

jzawadzki:
Hi,

regarding our discussion in #220 and #136 I'm sending first draft of checking if installer is up-to-date.
I have based my solution of how and when composer is checking it.
Please comment and tell if this solution is ok and what should I change to make it better!

Thanks,
JZ

### DIFF
--- a/src/Symfony/Installer/Application.php
+++ b/src/Symfony/Installer/Application.php
@@ -26,7 +26,7 @@ class Application extends ConsoleApplication
     {
         $commandName = $this->getCommandName($input);
 
-        if ($this->isInPharMode() && $commandName != 'self-update' && $commandName != 'selfupdate') {
+        if ($this->isInPharMode() && !in_array($commandName, array('self-update', 'selfupdate'), true)) {
             if (!$this->checkIfInstallerIsUpdated()) {
                 $output->writeln(
                     sprintf(
@@ -52,7 +52,7 @@ class Application extends ConsoleApplication
         $localVersion = $this->getVersion();
 
         if (false === $remoteVersion = @file_get_contents(self::VERSIONS_URL)) {
-            // as this is simple checking - we don't here care here if versions file is unavailable
+            // as this is simple checking - we don't care here if versions file is unavailable
             return true;
         }
 
--- a/src/Symfony/Installer/Application.php
+++ b/src/Symfony/Installer/Application.php
@@ -1,0 +1,66 @@
+<?php
+
+/*
+ * This file is part of the Symfony Installer package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Installer;
+
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Application as ConsoleApplication;
+
+/**
+ * This is main Symfony Installer console application class
+ *
+ */
+class Application extends ConsoleApplication
+{
+    const VERSIONS_URL = 'http://get.sensiolabs.org/symfony.version';
+
+    public function doRun(InputInterface $input, OutputInterface $output)
+    {
+        $commandName = $this->getCommandName($input);
+
+        if ($this->isInPharMode() && $commandName != 'self-update' && $commandName == 'selfupdate') {
+            if (!$this->checkIfInstallerIsUpdated()) {
+                $output->writeln(
+                    sprintf(
+                        " <comment>[WARNING]</comment> You are using outdated version of Symfony Installer.\n" .
+                        " It is recommended to update it by running \"%s selfupdate\" to get the latest version.",
+                        $_SERVER['PHP_SELF']
+                    )
+                );
+            }
+
+        }
+        return parent::doRun($input, $output);
+    }
+
+    public function isInPharMode()
+    {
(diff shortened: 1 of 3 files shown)


Pierstoval:
Hehe, we proposed the same feature at the same time :laughing: (#221)

src/Symfony/Installer/Application.php
```diff
@@ -0,0 +1,66 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Application as ConsoleApplication;
+
+/**
+ * This is main Symfony Installer console application class.
+ */
+class Application extends ConsoleApplication
+{
+    const VERSIONS_URL = 'http://get.sensiolabs.org/symfony.version';
```
stof: I suggest using ``get.symfony.com`` instead (this is the Symfony installer after all), as it also works (and even with https)

src/Symfony/Installer/Application.php
```diff
@@ -0,0 +1,66 @@
...
+
+    public function doRun(InputInterface $input, OutputInterface $output)
+    {
+        $commandName = $this->getCommandName($input);
+
+        if ($this->isInPharMode() && $commandName != 'self-update' && $commandName == 'selfupdate') {
```
stof: should be ``!= 'selfupdate'``

jzawadzki:
@Pierstoval yes - minutes away! :D  2 solutions are still better then no solution!
Good we have @javiereguiluz to decide ;)

src/Symfony/Installer/Application.php
jzawadzki: will change to `get.symfony.com` but we still don't use https here 
https://github.com/symfony/symfony-installer/blob/master/src/Symfony/Installer/SelfUpdateCommand.php#L125 ?
Is there any reason

stof: mostly because the last time we worked on the SelfUpdateCommand, SSL was not yet available

jzawadzki: updated to https - thanks for claryifing

src/Symfony/Installer/Application.php
```diff
@@ -0,0 +1,66 @@
...
+ */
+class Application extends ConsoleApplication
+{
+    const VERSIONS_URL = 'https://get.symfony.com/symfony.version';
+
+    public function doRun(InputInterface $input, OutputInterface $output)
+    {
+        $commandName = $this->getCommandName($input);
+
+        if ($this->isInPharMode() && $commandName != 'self-update' && $commandName != 'selfupdate') {
```
ogizanagi:
```diff
- if ($this->isInPharMode() && $commandName != 'self-update' && $commandName != 'selfupdate') {
+ if ($this->isInPharMode() && !in_array($commandName, array('self-update', 'selfupdate'), true) {
```

src/Symfony/Installer/Application.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        return 'phar://' === substr(__DIR__, 0, 7);
+    }
+
+    private function checkIfInstallerIsUpdated()
+    {
+        $localVersion = $this->getVersion();
+
+        if (false === $remoteVersion = @file_get_contents(self::VERSIONS_URL)) {
+            // as this is simple checking - we don't here care here if versions file is unavailable
```
ogizanagi:
```diff
- // as this is simple checking - we don't here care here if versions file is unavailable
+ // as this is simple checking - we don't care here if versions file is unavailable
```

(extra `here`)

src/Symfony/Installer/Application.php
jzawadzki: changed - thx for review

src/Symfony/Installer/Application.php
jzawadzki: fixed  - thank you!

src/Symfony/Installer/Application.php
```diff
@@ -0,0 +1,66 @@
...
+ */
+class Application extends ConsoleApplication
+{
+    const VERSIONS_URL = 'https://get.symfony.com/symfony.version';
+
+    public function doRun(InputInterface $input, OutputInterface $output)
+    {
+        $commandName = $this->getCommandName($input);
+
+        if ($this->isInPharMode() && !in_array($commandName, array('self-update', 'selfupdate'), true)) {
```
javiereguiluz: Just asking: shouldn't we also avoid to call this command for `list`, `help` and the other common actions? Maybe we could replace the `!in_array($commandName, array(...))` by `in_array($commandName, array('new', 'demo'))` ?

jzawadzki: At first I was :+1: You are right that those are mostly needed on `new` and `demo`, but after some considerations:

What if we change some command names/parameters and user will get old version on `help`  (different than e.g. on symfony.com site) - that could make some problems I think?

Maybe we should still check version there and set notification but not show it as warning?

ogizanagi: IMHO, commands arguments and options are not likely to change very often, and the `help` command will still be valid for the current installed version. If someone does not have a new option, I think he'll figure out his installer is outdated.

So :+1: about Javier suggestion.

jzawadzki: ok - updated!

javiereguiluz:
@jzawadzki thanks for this feature! (and thanks to @Pierstoval too for working on this feature in another pull request!)

PR #222, opened 2015-12-05
Check the installer version

Pierstoval:
This is a *basic* implementation that will compare the local version with the distant one.

Actually, this does the check on every command, but I'd like to discuss about how we could improve this, because I don't know what kind of implications there is in requesting the version on each command...

### DIFF
--- a/src/Symfony/Installer/Application.php
+++ b/src/Symfony/Installer/Application.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * This file is part of the Symfony Installer package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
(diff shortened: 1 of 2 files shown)


stof:
#221 has the advantage of skipping this warning for the self-update command itself

Pierstoval:
No problem for keeping #221 ;)

javiereguiluz:
I'm closing this in favor of #221. The other one was submitted earlier and the code looks a bit simpler. Anyway, thanks @Pierstoval for your help with this. As always, your work is really appreciated!

Pierstoval:
"Earlier" is just about some few minutes :laughing: But yeah I should have waited before submitting a PR for nothing :)
Thanks ;)

PR #223, opened 2015-12-06
Updated installation description

TheKnarf:
A modern development process involves relying upon a package manager
such as composer, and we can therefore use composers strengths as the
default method while still allowing for the old method as an alternative.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,37 +7,29 @@ full-stack framework.**
 Installing the installer
 ------------------------
 
-This step is only needed the first time you use the installer:
+To install `Symfony Installer` its recommended to first install [Composer](https://getcomposer.org/doc/00-intro.md#installation-linux-unix-osx). If you do not wish to use composer to manage the `Symfony Installer` you can check out the [alternative installation method](alternative-installation.md).
 
-### Linux and Mac OS X
+If you have composer installed you can install `Symfony Installer`:
 
-```bash
-$ sudo curl -LsS http://symfony.com/installer -o /usr/local/bin/symfony
-$ sudo chmod a+x /usr/local/bin/symfony
-```
-
-### Windows
 
 ```bash
-c:\> php -r "file_put_contents('symfony', file_get_contents('http://symfony.com/installer'));"
+$ composer require symfony/symfony-installer --dev
 ```
 
-Move the downloaded `symfony` file to your projects directory and execute
-it as follows:
+To allow you to call the command you eather have to use the command: 
 
 ```bash
-c:\> php symfony
+./vendor/bin/symfony
 ```
 
-If you prefer to create a global `symfony` command, execute the following:
+Or add the `./vendor/bin` to your path. On Linux and Mac OSX you can do this by the following command:
 
 ```bash
-c:\> (echo @ECHO OFF & echo php "%~dp0symfony" %*) > symfony.bat
+export PATH=$PATH./vendor/bin
 ```
 
-Then, move both files (`symfony` and `symfony.bat`) to any location included
-in your execution path. Now you can run the `symfony` command anywhere on your
-system.
+It's recommended to add this line to you `.bashrc` or `.zshrc` file so you don't have to run it every time you open a new terminal.
+
 
 Using the installer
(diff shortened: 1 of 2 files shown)


javiereguiluz:
@TheKnarf I agree about using Composer to manage project's dependencies. However, in this case I think Composer completely destroys the DX (developer experience) of the Symfony Installer. These are the reasons:

* It's much slower to install (the current installation method takes less than 1 second in a normal Internet connection).
* You are proposing to install the Installer for every project (because you don't install the package as a global Composer package). Meanwhile, the current installation is once-in-a-lifetime task.
* You require to edit the `.bashrc` or similar file, which doesn't work for Windows users and it's intimidating for lots of newcomers.
* The installer update becomes a bit weird, because you just install the Installer again, instead of executing the `self-update` command.

ogizanagi:
@javiereguiluz :+1: 

Also, in case you would suggest to install it globally, it implies potential conflicts with already globally installed packages through composer.

jzawadzki:
I agree with @javiereguiluz.
Downloading installer via composer isn't the best idea. Symfony installer repository is actually not to be used as 'package' as no other app I think would require it as dependency? It's mostly for generating PHAR from it. 

Installer was created to speed up the process (downloading pregenerated zip/tgz) and to introduct less experienced developers (maybe not so familiar with composer) to symfony. If someone knows how to use composer he can also not use Symfony Installer at all (using `composer create-project symfony/framework-standard-edition`).

@javiereguiluz btw I was thinking lately if we could/should add somewhere information about previous installation method by `composer create-project` that's still available? Just in case installer/symfony server have some problems to let users know that such method is still there?

javiereguiluz:
@jzawadzki I agree about showing the alternative installation method somehow. That's why a few weeks ago we added a link to the alternative download method in the Download page:

alternative_installation

jzawadzki:
@javiereguiluz :+1: great! thanks - it's good it's there:)

TheKnarf:
I belive that using composer should be the default option and not the alternative. Composer have become a very important part of the PHP ecosystem and its important that newcomers learns to use it as soon as possible.

@javiereguiluz you suggested that editing the `.bashrc` file would seem even more intimidating, and while I agree on that point you could still just call it with `./vendor/bin/symfony`, this will not require you to edit `.bashrc`.

@ogizanagi @javiereguiluz I agree that installing it globally is not an good idea. Theres a reason that composer have developer have dev-dependencies. I think that the `self-update` method should be removed or disabled from the composer install. This simplifies development as you will have to learn one less tool, you just use composer as usual. And you update dependencies as usual. 

@javiereguiluz Having to install it for every project is really a moot point as composer will cache the download. It might take some extra time the first time but would not take any more time the second turn.

Another reason for adding it as a dependency and using it as the de facto installation method is so that we can extend the functionality of `symfony installer` and adding a lot more value to it than whats possible with `composer create-project`. We could add possibility to customize which dependencies and bundles are shipped with an Symfony installation. Do you want to use Doctrine or Propel? Do you need the swiftmail bundle? What about phpunit vs phpspec?

_I think its time to embrace Composer._

javiereguiluz:
@TheKnarf rest assure that we love, use and recommend Composer. However, we use it for managing dependencies, not for creating projects.

I'm closing this issue as "won't fix", but please don't understand this as *"they think they are right and I'm wrong"*. It's not about being right or wrong. You proposal would ask us to change the Symfony Installation process radically. But we are very happy with it and most people are thrilled by the new installation method. That's why for now we are not planning to introduce big changes to it.

In any case, thanks for letting us your concerns about the installer and for engaging in a interesting discussion.

Pierstoval:
However, we can still add, in the docs, a proposal to require the installer globally:

```bash
$ composer global require symfony/installer
```

The update process would just be `composer global update` instead of `symfony selfupdate`, and in order to have it in the path, one just need to add `~/.composer/vendor/bin/` to their `PATH` or to add a symlink with `sudo ln -s ~/.composer/vendor/bin/symfony /usr/local/bin/symfony`.

PR #226, opened 2015-12-10
Intercept and customize error from Guzzle

emil-nasso:
When guzzle is initalized it tries to find a suitable handler to use. If no handler can be found (for example curl) an exception is thrown.

With this change we try to find the handler before creating the client so that we can pass the handler to the client manually.
Doing this, we are able to customize the error message to make it (potentially) less confusing for the users who might have no idea what guzzle is.

The issue was found when looking at PR #200

### DIFF
--- a/src/Symfony/Installer/DownloadCommand.php
+++ b/src/Symfony/Installer/DownloadCommand.php
@@ -19,6 +19,7 @@ use Distill\Strategy\MinimumSize;
 use GuzzleHttp\Client;
 use GuzzleHttp\Exception\ClientException;
 use GuzzleHttp\Event\ProgressEvent;
+use GuzzleHttp\Utils;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\InputInterface;
@@ -221,7 +222,15 @@ abstract class DownloadCommand extends Command
             $defaults['debug'] = true;
         }
 
-        return new Client(array('defaults' => $defaults));
+        try {
+            $handler = Utils::getDefaultHandler();
+        } catch (\RuntimeException $e) {
+            throw new \RuntimeException(
+                'The Symfony installer requires the php-curl extension or the allow_url_fopen ini setting.'
+            );
+        }
+
+        return new Client(array('defaults' => $defaults, 'handler' => $handler));
     }
 
     /**


src/Symfony/Installer/DownloadCommand.php
```diff
@@ -221,7 +222,15 @@ protected function getGuzzleClient()
             $defaults['debug'] = true;
         }
 
-        return new Client(array('defaults' => $defaults));
+        try {
+            $handler = Utils::getDefaultHandler();
+        } catch (\RuntimeException $e) {
+            throw new \RuntimeException(
+                'The symfony installer requires the php-curl extension or the '
```
xabbuh: Symfony

emil-nasso: @xabbuh fixed

src/Symfony/Installer/DownloadCommand.php
```diff
@@ -221,7 +222,15 @@ protected function getGuzzleClient()
             $defaults['debug'] = true;
         }
 
-        return new Client(array('defaults' => $defaults));
+        try {
+            $handler = Utils::getDefaultHandler();
+        } catch (\RuntimeException $e) {
+            throw new \RuntimeException(
+                'The Symfony installer requires the php-curl extension or the '
```
Pierstoval: I don't think wrapping this message is the best idea for code style :)

src/Symfony/Installer/DownloadCommand.php
```diff
@@ -221,7 +222,14 @@ protected function getGuzzleClient()
             $defaults['debug'] = true;
         }
 
-        return new Client(array('defaults' => $defaults));
+        try {
+            $handler = Utils::getDefaultHandler();
+        } catch (\RuntimeException $e) {
+            throw new \RuntimeException(
+                'The Symfony installer requires the php-curl extension or the allow_url_fopen ini setting.');
```
ogizanagi: `);` should be on the next line, or keep everything inlined with the `\RuntimeException` constructor call.

emil-nasso: fixed

Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM debian:jessie
+VOLUME .:/root/shared
+WORKDIR /root/shared
+RUN apt-get update && apt-get install -y php5-cli
+
+ENTRYPOINT /bin/bash
```
Pierstoval: Is this really needed to have a dockerfile in this repo actually?

emil-nasso: Wow. I can't do this right.... I accidentally committed it. Really sorry...

javiereguiluz:
@emil-nasso I love it when an application displays meaningful error messages. Thanks for helping us improving the Installer error messages.

PR #227, opened 2015-12-10
Update out of date lock file

victorhaggqvist:
This fixes the out of date lock file in the v1.5.0 release


javiereguiluz:
@victorhaggqvist thanks for fixing this issue.

xabbuh:
@javiereguiluz What is the process when you draft a new release? Is there a way we can reduce the risk of an outdated `composer.lock` file (it's not a big issue but if we can avoid this that would be even better)?

PR #232, opened 2016-01-03
Symfony install review

94noni:
Mini review + fix for version

:information_source: Could not test (and correct the test) as the .phar could not be created.
I have to check why the build fail.

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -83,16 +83,17 @@ class NewCommand extends DownloadCommand
 
     /**
      * Checks whether the given Symfony version is installable by the installer.
+     * (see http://symfony.com/roadmap)
      * Due to the changes introduced in the Icu/Intl components
      * (see http://symfony.com/blog/new-in-symfony-2-6-farewell-to-icu-component)
      * not all the previous Symfony versions are installable by the installer.
      *
      * The rules to decide if the version is installable are as follows:
      *
-     *   - 2.0, 2.1, 2.2 and 2.4 cannot be installed because they are unmaintained.
+     *   - 2.0, 2.1, 2.2, 2.4, 2.5 and 2.6 cannot be installed because they are unmaintained.
      *   - 2.3 can be installed starting from version 2.3.21 (inclusive)
-     *   - 2.5 can be installed starting from version 2.5.6 (inclusive)
-     *   - 2.6, 2.7, 2.8 and 2.9 can be installed regardless the version.
+     *   - 2.7 and 2.8 can be installed regardless the version.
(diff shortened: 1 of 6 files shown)


src/Symfony/Installer/NewCommand.php
```diff
@@ -133,8 +134,8 @@ protected function checkSymfonyVersionIsInstallable()
             }
         }
 
-        // 2.0, 2.1, 2.2 and 2.4 cannot be installed because they are unmaintained
-        if (preg_match('/^2\.[0124]\.\d{1,2}$/', $this->version)) {
+        // 2.0, 2.1, 2.2, 2.4, 2.5 and 2.6 cannot be installed because they are unmaintained
```
javiereguiluz: The change is wrong. Starting from 2.5.0, all Symfony versions are installable.

src/Symfony/Installer/NewCommand.php
```diff
@@ -83,16 +83,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
      * Checks whether the given Symfony version is installable by the installer.
+     * (see http://symfony.com/roadmap)
      * Due to the changes introduced in the Icu/Intl components
      * (see http://symfony.com/blog/new-in-symfony-2-6-farewell-to-icu-component)
      * not all the previous Symfony versions are installable by the installer.
      *
      * The rules to decide if the version is installable are as follows:
      *
-     *   - 2.0, 2.1, 2.2 and 2.4 cannot be installed because they are unmaintained.
+     *   - 2.0, 2.1, 2.2, 2.4, 2.5 and 2.6 cannot be installed because they are unmaintained.
```
javiereguiluz: Same comment as https://github.com/symfony/symfony-installer/pull/232/files#r48697441

src/Symfony/Installer/NewCommand.php
94noni: @javiereguiluz 

yes indeed

but reading http://symfony.com/roadmap?version=2.5#checker + http://symfony.com/roadmap?version=2.6#checker,
I thought it will be a good idea to only allow new dev to install maintained version :)

94noni:
Could not test (and correct test) as the .phar could not be created.
I have to check why the build fail.

src/Symfony/Installer/NewCommand.php
javiereguiluz: Although it generally makes no sense to install a deprecated version, the installer should allow to install any past version which is compatible with the installer.

javiereguiluz:
@94noni I'd like to merge these changes, so I've created #246 to finish your work. I just reverted the changes I mentioned you in the NewCommand. I used your original commit, so you'll get full credit for your work. Thanks!

PR #236, opened 2016-01-30
remove pre-configured platform PHP version

### LINKED ISSUE #234: Remove pre-configured platform PHP version
Starting with symfony/symfony-standard@d0e9308e807aae02c332897d610ad5eb49d81c1e, the Symfony Standard Edition will define the platform PHP version as 5.3.3 to avoid issues for people using older PHP versions than the system on which the downloadable packages are built. However, this will lead to weird behaviour for people who actually use newer PHP versions and want to install packages that are incompatible with PHP 5.3. So I think we should simply remove the platform PHP version configuration from the `composer.json` file after the installation has finished.

xabbuh:
This resolves #234.

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -350,6 +350,18 @@ class NewCommand extends DownloadCommand
             unset($contents['description']);
         }
 
+        if (isset($contents['config']['platform']['php'])) {
+            unset($contents['config']['platform']['php']);
+
+            if (empty($contents['config']['platform'])) {
+                unset($contents['config']['platform']);
+            }
+
+            if (empty($contents['config'])) {
+                unset($contents['config']);
+            }
+        }
+
         if (isset($contents['extra']['branch-alias'])) {
             unset($contents['extra']['branch-alias']);
         }


WouterJ:
:+1:

src/Symfony/Installer/NewCommand.php
```diff
@@ -350,6 +350,18 @@ protected function updateComposerJson()
...
         }
 
+        if (isset($contents['config']['platform']['php'])) {
+            unset($contents['config']['platform']['php']);
+
+            if (empty($contents['config']['platform'])) {
+                unset($contents['config']['platform']);
+            }
+
+            if (empty($contents['config'])) {
```
bocharsky-bw: What about to make it nested to the previous if statement? The `$contents['config']` could be empty only when we unset `$contents['config']['platform']`, otherwise it probably has at least one item.

```php
            if (empty($contents['config']['platform'])) {
                unset($contents['config']['platform']);

                if (empty($contents['config'])) {
                    unset($contents['config']);
                }
            }
```

xabbuh:
ping @symfony/deciders 

This seems to cause issues for people not familiar with the platform config in Composer (see https://github.com/composer/composer/issues/4319#issuecomment-180273853)

src/Symfony/Installer/NewCommand.php
xabbuh: I wanted to avoid nesting control structures too deeply.

bocharsky-bw: What purpose to do it? Only to reduce complexity?

xabbuh: Yes, but I am happy to change it if others feel uncomfortable with how it is right now too.

bocharsky-bw: OK, let's wait for others a bit, but I think I'm fine to keep it as is

Pierstoval: I think this does not complexify the behavior, it just simplifies the outputted JSON

Tobion:
Does this result in a warning that "composer.json is not in sync with lock file"? Or is this config option not considered for this sync check by composer?

xabbuh:
The lock file is synchronised after the `composer.json` file has been changed: https://github.com/xabbuh/symfony-installer/blob/issue-234/src/Symfony/Installer/NewCommand.php#L371

Tobion:
LGTM :+1:

Pierstoval:
:+1:

Seldaek:
Just FYI https://github.com/composer/composer/pull/4881 should solve the confusion side.. I'd argue having platform config is a good thing but obviously if people don't take care of setting it correctly it's a bad idea, so maybe not having one by default is better.

stof:
@Seldaek but having it set to 5.3 in new Symfony 2.8 projects is not a good idea. symfony standard-edition needs this value because of its min supported version, but I hope new projects actually target higher PHP runtimes when deploying.
So keeping the Symfony setting is not a good idea. People needing this setting would need to configure it according to their need anyway. and relying on the current runtime is a better behavior for people not aware of this composer feature.

So :+1: for this change

xabbuh:
@javiereguiluz What do you think?

Pierstoval:
I think it can be merged now and released for newcomers to avoid this issue :+1:

dunglas:
:+1:

javiereguiluz:
@xabbuh I'm sorry it took me so long to merge your contribution. It's merged now.

PR #238, opened 2016-02-17
box build fails with invalid path error

smatyas:
This PR fixes the issue when the `box build` command fails with the following error:

```
$ box build
Building...


  [BadMethodCallException]
  Entry src//Symfony/Installer/AboutCommand.php does not exist and cannot be created: phar error: invalid path "src//Symfony/Installer/AboutCommand.php" contains double slash


build [-c|--configuration CONFIGURATION]
```

The [examples in the box wiki](https://github.com/box-project/box2/wiki/Composer) also uses directory names without the trailing slash.

Tested on this system config:
php: 5.5.29
box: 2.6.1
os: OS X 10.10.5

### DIFF
--- a/box.json
+++ b/box.json
@@ -1,5 +1,5 @@
 {
-    "directories": ["src/"],
+    "directories": ["src"],
     "finder": [
         {
             "name": "*.php",


Pierstoval:
Can't Box itself solve this issue by removing any trailing slash?

smatyas:
Thanks @Pierstoval, let's see what's their opinion. I created a PR there as well.

xabbuh:
Is this still an issue given that box-project/box2#130 was merged?

smatyas:
Hi,
just tested using box v2.7.0, it's working as expected now.

Issue #242, opened 2016-03-18
Empty .gitignore on symfony new project


mariuspuiu:
New projects created via the installer have empty `.gitignore` file in the root of the app by default.

![image](https://cloud.githubusercontent.com/assets/878197/13881511/db6509a8-ed29-11e5-9509-488d175d2c09.png)

Meaning that someone might add by mistake the `cache`, `logs` and others to the repository:

![image](https://cloud.githubusercontent.com/assets/878197/13879578/dc943924-ed20-11e5-981c-c0e4f18f36c5.png)

Pierstoval:
According to your `symfony new my_project` you should have run `ls my_project/.gitignore`, and it would have worked actually :confused: 

I just tested on my machine and it works.

mariuspuiu:
Sorry, forgot to cd to my_project. Still, the .gitignore file is empty.

mariuspuiu:
It seems that I have an issue on my system, I removed the `@` in front of `file_get_contents()` (`DownloadCommand.php`, line `326`) and the output is:
```
PHP Warning:  file_get_contents(https://raw.githubusercontent.com/symfony/symfony-standard/v3.0.3/.gitignore): failed to open stream: Connection timed out in /var/www/symfony-installer/src/Symfony/Installer/DownloadCommand.php on line 326
```

With cURL, it works:
```
curl https://raw.githubusercontent.com/symfony/symfony-standard/v3.0.3/.gitignore
/app/config/parameters.yml
/build/
/phpunit.xml
/var/*
!/var/cache
/var/cache/*
!var/cache/.gitkeep
!/var/logs
/var/logs/*
!var/logs/.gitkeep
!/var/sessions
/var/sessions/*
!var/sessions/.gitkeep
!var/SymfonyRequirements.php
/vendor/
/web/bundles/
```

Pierstoval:
Check your PHP configuration:
* Do you have `openssl` module installed and enabled?
* Is `http` registered as a stream?
* Which version do you use?

mariuspuiu:
```
Registered Stream Socket Transports	tcp, udp, unix, udg, ssl, sslv3, sslv2, tls, tlsv1.0, tlsv1.1, tlsv1.2
```
```
OpenSSL support enabled
OpenSSL Library Version OpenSSL 1.0.1e-fips 11 Feb 2013
```

brightdroid:
@mariuspuiu are you also behind a proxy? Then it's definitely the same problem as #243.

[...]

PR #244, opened 2016-03-21
Issue #242 - Empty .gitignore file

mariuspuiu:
`createGitIgnore()` function had the wrong path for the `.gitignore` file download. Removed the `v` from the URL.


Pierstoval:
Which file causes an error?

Because https://raw.githubusercontent.com/symfony/symfony-standard/v2.8.0/.gitignore works very well :confused:

PR #246, opened 2016-04-03
Updated the Symfony versions displayed in README and commands

javiereguiluz:
This supersedes #232.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2014-2015 Fabien Potencier
+Copyright (c) 2014-2016 Fabien Potencier
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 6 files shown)


stof:
@javiereguiluz can you fix the testsuite ?

javiereguiluz:
The error is caused by the infamous Symfony error that left the `-DEV` suffix in the version number (https://github.com/symfony/symfony/issues/18390 & related). What's the best approach to solve this in our test suite?

javiereguiluz:
Let's merge this as is and then we'll think about how to fix our test suite. Thanks!

xabbuh:
Is it worth to change anything? The failure is not related to the installer but to the built packages and will be good again with the next release.

javiereguiluz:
@xabbuh I think we should fix it because waiting a month is too much and the temporary fix is trivial: https://github.com/symfony/symfony-installer/pull/248

PR #247, opened 2016-04-04
Made all network connections proxy-compatible

### LINKED ISSUE #243: proxy is not used everywhere
In some places the installer still use "file_get_contents", which does not work behind a proxy.
e.g.
- DownloadCommand::createGitIgnore
- SelfUpdateCommand...
- ...

javiereguiluz:
This fixes #243.

### DIFF
--- a/src/Symfony/Installer/SelfUpdateCommand.php
+++ b/src/Symfony/Installer/SelfUpdateCommand.php
@@ -27,15 +27,12 @@ use Symfony\Component\Filesystem\Exception\IOException;
  * @author Stephane PY <py.stephane1@gmail.com>
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
  */
-class SelfUpdateCommand extends Command
+class SelfUpdateCommand extends DownloadCommand
 {
-    /** @var Filesystem */
-    private $fs;
-
-    /** @var OutputInterface */
-    private $output;
-
     private $tempDir;
+
+    /** @var string */
+    private $latestInstallerVersion;
 
     /** @var string  the URL where the latest installer version can be downloaded */
     private $remoteInstallerFile;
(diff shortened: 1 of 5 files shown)


src/Symfony/Installer/SelfUpdateCommand.php
```diff
@@ -98,16 +97,16 @@ protected function execute(InputInterface $input, OutputInterface $output)
                 ->cleanUp()
             ;
         } catch (IOException $e) {
+            if ($output->isVeryVerbose()) {
+                echo $e->getMessage();
```
Pierstoval: You could just add the `$e` var as third argument of the `RuntimeException` thrown right after it

stof: this should use the output btw, not a straight ``echo``

src/Symfony/Installer/SelfUpdateCommand.php
```diff
@@ -85,8 +82,10 @@ protected function initialize(InputInterface $input, OutputInterface $output)
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        if ($this->installerIsUpdated()) {
-            return;
+        if ($this->isInstallerUpdated()) {
+            $this->output->writeln(sprintf('// Symfony Installer is <info>already updated</info> to the latest version (%s).', $this->latestInstallerVersion));
```
stof: missing return after that, to avoid downloading it

stof:
Why is the refactoring of the installer version check shipped in the same PR (without being described in the PR) ?

javiereguiluz:
@stof I agree that's a bad practice and I'm sorry about it.

stof:
@javiereguiluz note that this hidden change is even the one preventing to merge this PR, as it is the buggy part...

PR #248, opened 2016-04-06
Temporary fix for failing tests

javiereguiluz:
Symfony released some stable versions with the `-DEV` suffix in their versions. This broke our tests. Let's fix this temporarily until the next maintenance versions are released next month.


xabbuh:
fair enough :+1:

PR #252, opened 2016-06-02
Add missing DocBlock.

regniblod:


src/Symfony/Installer/AboutCommand.php
```diff
@@ -22,15 +22,26 @@
  */
 class AboutCommand extends Command
 {
+    /**
+     * @var string The current version of the Symfony installer
```
xabbuh: This is actually not needed as the type can be inferred from the constructor.

javiereguiluz: You are right ... but let's leave it for consistency sake.

javiereguiluz:
@regniblod thanks for taking care of updating/finishing all the PHPdoc blocks!

PR #256, opened 2016-06-06
Remove the "config.platform.php" option when installing the Symfony Demo app

### LINKED ISSUE #255: Remove the config.platform.php config after installing the Symfony Demo app
The `NewCommand` removes the `config.platform.php` Composer option after installing Symfony (look in the `updateComposerJson()` method).

We should do the same for the `DemoCommand` to avoid this problem when installing the Symfony Demo app.

javiereguiluz:
This fixes #255.

### DIFF
--- a/src/Symfony/Installer/DemoCommand.php
+++ b/src/Symfony/Installer/DemoCommand.php
@@ -156,33 +156,6 @@ class DemoCommand extends DownloadCommand
     }
 
     /**
-     * Updates the composer.json file to provide better values for some of the
-     * default configuration values.
-     *
-     * @return $this
-     */
-    protected function updateComposerJson()
-    {
-        $composerConfig = $this->getProjectComposerConfig();
-
-        if (isset($composerConfig['config']['platform']['php'])) {
-            unset($composerConfig['config']['platform']['php']);
-
-            if (empty($composerConfig['config']['platform'])) {
-                unset($composerConfig['config']['platform']);
-            }
-
-            if (empty($composerConfig['config'])) {
-                unset($composerConfig['config']);
-            }
-        }
-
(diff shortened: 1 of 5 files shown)


src/Symfony/Installer/DownloadCommand.php
```diff
@@ -545,6 +545,107 @@ protected function getUrlContents($url)
...
+     * It saves the given PHP array as the project's Composer config. In addition
+     * to JSON-serializing the contents, it synchronizes the composer.lock file to
+     * avoid out-of-sync Composer errors.
+     *
+     * @param array $config
+     */
+    protected function saveProjectComposerConfig(array $config)
+    {
+        $composerJsonFilepath = $this->projectDir.'/composer.json';
+        $this->fs->dumpFile($composerJsonFilepath, json_encode($config, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES)."\n");
```
Pierstoval: These JSON constants are available starting with PHP 5.4, so 5.3 users will get an error.
 
Maybe use the int flag manually : `128 | 64` and add a comment about the constants?

src/Symfony/Installer/DemoCommand.php
```diff
@@ -155,6 +156,33 @@ private function displayInstallationResult()
     }
 
     /**
+     * Updates the composer.json file to provide better values for some of the
+     * default configuration values.
+     *
+     * @return $this
+     */
+    protected function updateComposerJson()
```
xabbuh: Why not `private`?

javiereguiluz: Good question! I've refactored everything to move this method to the abstract parent class.

javiereguiluz:
This is now ready for the final review. Thanks!

PR #257, opened 2016-06-06
Check that the system has the PHP version required by Symfony

### LINKED ISSUE #239: No warning/error given when installing Symfony 3.0 with PHP 5.4
Running

    $ symfony new my_project_name

as [suggested by the installation guide](http://symfony.com/doc/3.0/book/installation.html#creating-the-symfony-application) with PHP 5.4, will install Symfony 3.0, but issue no warnings or errors about the installation being incompatible with the PHP version used.  As documented in the [reference for Symfony 3.0](http://symfony.com/doc/3.0/reference/requirements.html) the minimum required version of PHP for Symfony 3.0 is 5.5.9.

This leads to a bad development experience as next thing suggested to do by the guide is to run the PHP development web server:

    $ php bin/console server:run
    PHP Parse error:  syntax error, unexpected 'class' (T_CLASS), expecting identifier (T_STRING) or variable (T_VARIABLE) or '{' or '$' in vendor/symfony/symfony/src/Symfony/Component/Debug/ErrorHandler.php on line 386

javiereguiluz:
This fixes #239.

### DIFF
--- a/tests/Symfony/Installer/Tests/IntegrationTest.php
+++ b/tests/Symfony/Installer/Tests/IntegrationTest.php
@@ -64,8 +64,12 @@ class IntegrationTest extends \PHPUnit_Framework_TestCase
     /**
      * @dataProvider provideSymfonyInstallationData
      */
-    public function testSymfonyInstallation($versionToInstall, $messageRegexp, $versionRegexp)
+    public function testSymfonyInstallation($versionToInstall, $messageRegexp, $versionRegexp, $minPhpVersion)
     {
+        if (version_compare(phpversion(), $minPhpVersion, '<')) {
+            $this->markTestSkipped(sprintf('This test requires PHP %s or higher.', $minPhpVersion));
+        }
+
         $projectDir = sprintf('%s/my_test_project', sys_get_temp_dir());
         $this->fs->remove($projectDir);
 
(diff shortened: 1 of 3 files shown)


xabbuh:
You will now have to skip some tests when on PHP 5.4.

javiereguiluz:
@xabbuh yes. I did that in a80196a ... but there are other failing tests which I can't fix. I'll try to fix them tomorrow.

javiereguiluz:
This is now ready for the final review. Thanks!

tests/Symfony/Installer/Tests/IntegrationTest.php
```diff
@@ -77,6 +81,19 @@ public function testSymfonyInstallation($versionToInstall, $messageRegexp, $vers
         $this->assertRegExp($versionRegexp, $output);
     }
 
+    /**
+     * @expectedException \RuntimeException
+     * @expectedExceptionMessageRegExp /.+The selected version \(3.0.0\) cannot be installed because it requires.+PHP 5.5.9 or higher and your system has PHP 5.4.* installed.+/s
+     */
+    public function testSymfonyRequiresNewerPhpVersion()
+    {
+        if (version_compare(phpversion(), '5.5.0', '>=')) {
```
xabbuh: `PHP_VERSION_ID >= 55000`

xabbuh:
Should we make it possible to skip the version check to be able to bootstrap a project even on a system that isn't fully ready to run the application itself?

tests/Symfony/Installer/Tests/IntegrationTest.php
javiereguiluz: Done. Thanks.

javiereguiluz:
@xabbuh I'd prefer to keep failing if your system is not ready to run the Symfony version you want to install.

javiereguiluz:
I think this needs more work. If you don't pass any version, you are using `latest`, which bypasses this new check, although it also requires 5.5.9.

javiereguiluz:
This is now ready for a new review. I changed everything about how we decide if a version is installable.

PR #261, opened 2016-06-25
Fixed the SymfonyRequirements check for the demo application

javiereguiluz:
The Symfony Demo was upgraded to Symfony 3 yesterday. If you install it, you get this error:

symfony_demo_error

The problem is that the `isSymfony3()` method is not suitable for the demo application. We cannot improve that method to get the Symfony version via the `--version` option because at that point we are about to check the Symfony Requirements to see if the computer can run the demo application.


Pierstoval:
For the `isSymfony3()` can't we use an env var?

Issue #266, opened 2016-07-12
Self update is not detecting the installed version


javiereguiluz:
A Symfony user tweeted about a possible error in the self-update command: https://twitter.com/fabricebaumann/status/752844734717198337

This week I'm too busy to verify or fix the error. Any help here would be greatly appreciated. Thanks!

javiereguiluz:
Closing it as fixed by #269.

PR #269, opened 2016-07-15
#266 Self update is not detecting the installed version

asvinours:
When the self-update command is initialized, the current version and latest version of the installer are not initialized.
The code is currently comparing NULL with NULL, which returns true.

By adding the parent::initialize() call, the DownloadCommand set the current and latest version of the installer, and then the version_compare function is actually really comparing the real versions.

### DIFF
--- a/src/Symfony/Installer/DownloadCommand.php
+++ b/src/Symfony/Installer/DownloadCommand.php
@@ -65,12 +65,12 @@ abstract class DownloadCommand extends Command
     /**
      * @var string The latest installer version
      */
-    private $latestInstallerVersion;
+    protected $latestInstallerVersion;
 
     /**
      * @var string The version of the local installer being executed
      */
-    private $localInstallerVersion;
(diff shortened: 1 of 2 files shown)


Archmonster:
Was about to post a reply to the bug report, but since you've already submitted a pull request i'm going to reply to you.

While it is true that the issue is version_compare not receiving the correct values, it does not solve the underlying issue, which is storing version numbers in private properties of DownloadCommand, and then trying to use / override them from a subclass. Both $latestInstallerVersion and $localInstallerVersion should be made protected instead of private and if you rely on parent::initialize() to set their values you should remove the duplicated code from SelfUpdateCommand::initialize().

asvinours:
I have updated my pull request to remove duplicated code and change the scope of the attributes of the DownloadCommand.

javiereguiluz:
Fabrice, thank you very much for reporting this error in the first place and then, for providing the pull request to fix it. Thanks!

PR #272, opened 2016-07-22
Added an option to force the installer update

javiereguiluz:
In the past, I introduced several bugs that prevented to update the installer: first it was a server-side issue with version numbers ... and then it was a client-side issue while comparing versions (see #266). The only solution in those cases is to re-install the installer 😢 

To avoid further issues, I propose to add a `--force-update (-f)` option that updates the local installer to whichever version is published by Symfony, without checking or comparing any version.


xabbuh:
👍

PR #276, opened 2016-09-02
Fixed some issues for the self-update command

smatyas:
Currently, the `self-update` command does not return an alerting enough exit message or status code on permission errors.
I saw in the code that it was intended, but it did not work well enough.

This PR fixes this issue by modifying the thrown exception type in certain cases, and by making the fallback exception to return error code 1, and by coloring the rollback message.

Before:
image

After:
image


javiereguiluz:
@smatyas it took us some time ... but this is finally merged! thanks for working on it.

PR #278, opened 2016-09-30
Fix lowercase issues with unicode characters

javiereguiluz:
@Pierstoval I tried your proposal in #197 but it didn't work for me. This new PR proposes to fix most of the errors for European languages. Sadly it won't work for Japanese, Chinese, Arabic, Hebrew, etc.


stof:
I would also remove any char not allowed in a composer package name (see the composer source code for their regex)

tests/Symfony/Installer/Tests/NewCommandTest.php
```diff
@@ -0,0 +1,30 @@
...
+
+    public function getProjectNames()
+    {
+        return [
+            ['foobar', 'foobar'],
+            ['áèîøūñ', 'aeioun'],
+            ['çįßłŵž', 'cislwz'],
+        ];
+    }
+}
```
stof: missing EOL

Pierstoval:
👍 for this 😄 Just need to make sure the box build works for php 5.4, or remove the travis build for this case

PR #279, opened 2016-10-06
Extracted the code related to Composer into a new ComposerManager class

javiereguiluz:
* This is just a code refactor, no new features were introduced.
* This is the first step into improving the code quality and making it easier to test.

---

I have a question: I don't know which is the best way to uncomment [these lines](https://github.com/symfony/symfony-installer/compare/master...javiereguiluz:refactor_composer?expand=1#diff-32356c397bce66f4bdab4408a6c6ef72R92). Should we pass the `output` object to the Composer manager? Thanks.

### DIFF
--- a/src/Symfony/Installer/Manager/ComposerManager.php
+++ b/src/Symfony/Installer/Manager/ComposerManager.php
@@ -1,0 +1,193 @@
+<?php
+
+namespace Symfony\Installer\Manager;
+
+use Symfony\Component\Filesystem\Filesystem;
+
+class ComposerManager
+{
+    private $projectDir;
+    private $fs;
+
+    public function __construct($projectDir)
+    {
+        $this->projectDir = $projectDir;
+        $this->fs = new Filesystem();
+    }
+
+    public function initializeProjectConfig()
+    {
+        $composerConfig = $this->getProjectConfig();
+
+        if (isset($composerConfig['config']['platform']['php'])) {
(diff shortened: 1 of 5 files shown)


src/Symfony/Installer/Manager/ComposerManager.php
```diff
@@ -0,0 +1,202 @@
...
+        if (!is_writable($composerJsonPath)) {
+//            if ($this->output->isVerbose()) {
+//                $this->output->writeln(sprintf(
+//                    " <comment>[WARNING]</comment> Project's Composer config cannot be updated because\n".
+//                    " the <comment>%s</comment> file is not writable.\n",
+//                    $composerJsonPath
+//                ));
+//            }
+
+            return $this;
```
stof: returning itself looks wrong to me.

And the fact that this is extracted into a dedicated private method does not stop the process anymore when the path is not writable (well, it will likely stop it due to the broken return value)

javiereguiluz: Yes, the `return $this;` was a leftover of the previous code.

In any case, we can remove the old warning message because returning an empty array when the Composer config is not readable will make the application work as expected.

javiereguiluz:
Closing in favor of #282 because there are some errors that I can't fix. @stof I did the changes that you asked. Thanks.

PR #284, opened 2016-10-10
Use 'server:start' instead of 'server:run' when possible

javiereguiluz:
`server:start` is a better option ... but it's not always possible to run it.

### DIFF
--- a/src/Symfony/Installer/DemoCommand.php
+++ b/src/Symfony/Installer/DemoCommand.php
@@ -145,7 +145,7 @@ class DemoCommand extends DownloadCommand
             ));
         }
 
-        $serverRunCommand = extension_loaded('pcntl') ? 'server:start' : 'server:run';
+        $serverRunCommand = version_compare($this->version, '2.6.0', '>=') && extension_loaded('pcntl') ? 'server:start' : 'server:run';
 
         $this->output->writeln(sprintf(
             "    1. Change your current directory to <comment>%s</comment>\n\n".
--- a/src/Symfony/Installer/DemoCommand.php
+++ b/src/Symfony/Installer/DemoCommand.php
@@ -145,7 +145,7 @@ class DemoCommand extends DownloadCommand
             ));
         }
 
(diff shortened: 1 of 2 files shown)


lyrixx:
Should be deal with Symfony Version also ? because this command does not exist in older SF version.

javiereguiluz:
@lyrixx you are right! I've improved the condition to check that we're using Sf 2.6.0 or higher. Thanks!

src/Symfony/Installer/DemoCommand.php
```diff
@@ -145,11 +145,13 @@ private function displayInstallationResult()
             ));
         }
 
+        $serverRunCommand = version_compare($this->version, '2.6.0', '>=') && extension_loaded('pcntl') ? 'server:start' : 'server:run';
```
xabbuh: Do we need this for the demo application?

src/Symfony/Installer/NewCommand.php
```diff
@@ -256,14 +256,15 @@ protected function displayInstallationResult()
         }
 
         $consoleDir = ($this->isSymfony3() ? 'bin' : 'app');
+        $serverRunCommand = version_compare($this->version, '2.6.0', '>=') && extension_loaded('pcntl') ? 'server:start' : 'server:run';
```
xabbuh: You also need to check for HHVM.

src/Symfony/Installer/DemoCommand.php
javiereguiluz: I don't know why should it be different for the demo app 😕

xabbuh: You cannot select a version of the demo application, can you? And the one that is going to be installed is using Symfony 3 already, isn't it?

javiereguiluz: You are so right!! I'm sorry for this mistake. It's now fixed. Thanks!

PR #285, opened 2016-10-30
Simplify code

fabpot:
Unless I'm missing something, this code always returns tgz.

### DIFF
--- a/src/Symfony/Installer/DownloadCommand.php
+++ b/src/Symfony/Installer/DownloadCommand.php
@@ -129,14 +129,7 @@ abstract class DownloadCommand extends Command
     {
         $this->output->writeln(sprintf("\n Downloading %s...\n", $this->getDownloadedApplicationType()));
 
-        // decide which is the best compressed version to download
-        $distill = new Distill();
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Indeed it looks like it's useless ... but this code chooses the minimum size file compatible with your operating system. So it chooses `.zip`when using the installer on Windows systems.

Our original idea was that someday Symfony would publish for example `.bz2` packages and then we'd add that format to the list.

fabpot:
If you add support to `bz2`, then hardcode `bz2` as it's going to always be returned. Are you sure about zip on Windows?

jzawadzki:
I've tested it just now on `Win10/php 7.0.12` and Installer has downloaded `.tgz`

xabbuh:
IIRC the behaviour depends on your Windows installation (i.e. version of the OS and which additional tools are installed). So on many systems the gzipped tar would still be used, but the code as is allows Distill to fall back to zip for some Windows systems.

javiereguiluz:
Let's close this old PR because the ZIP file is downloaded under some circumstances on Windows.

PR #302, opened 2017-12-26
Display a warning message when trying to install Symfony 4.x versions

### LINKED ISSUE #300: Gentle deprecation of this tool
This idea was proposed by @weaverryan on the Symfony Slack chat:

-----

This installer is deprecated starting from Symfony 4. We could do a last release of the installer to default to 3.4 and print a big warning message when trying to install Symfony 4.

Currently `symfony new blah` fails with `The selected version (4.0.1) cannot be installed...`

javiereguiluz:
This fixes #300.

### DIFF
--- a/src/Symfony/Installer/NewCommand.php
+++ b/src/Symfony/Installer/NewCommand.php
@@ -139,7 +139,7 @@ class NewCommand extends DownloadCommand
                 throw new \RuntimeException(sprintf(
                     "The selected branch (%s) does not exist, or is not maintained.\n".
                     "To solve this issue, install Symfony with the latest stable release:\n\n".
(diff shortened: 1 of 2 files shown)


src/Symfony/Installer/NewCommand.php
```diff
@@ -329,12 +339,12 @@ protected function updateParameters()
     protected function updateComposerConfig()
     {
         parent::updateComposerConfig();
-        $this->composerManager->updateProjectConfig([
+        $this->composerManager->updateProjectConfig(array(
```
Pierstoval: No need for this change, the installer is compatible with 5.4+ only

PR #303, opened 2017-12-31
Updated the README to no longer recommend using the Symfony Installer

javiereguiluz:
Just a proposal. Maybe it's too early for this deprecation?


README.md
```diff
@@ -1,157 +1,15 @@
...
-# Windows
-c:\> php symfony new my_project 2.8.1
-```
-
-**5. Install the Symfony demo application**
-
-The Symfony Demo is a reference application developed using the official Symfony
-Best Practices:
+**The Symfony Installer is no longer recommended to start new projects based on
+the Symfony full-stack framework.** Instead, use [Composer](https://getcomposer.org/):
```
Pierstoval: I would instead redirect to the "Download" section of symfony.com website. Saying "use composer" does not explain how to do use Composer. The website has a clear explanation.

README.md
```diff
@@ -1,157 +1,15 @@
...
-
-```ini
-; make sure that the following line is uncommented
-extension=php_openssl.dll
-```
-
-2) Check that the path to the file that contains the security certificates
-exists and is defined in `php.ini`:
+# Symfony 2.x and 3.x
+$ composer create-project symfony/framework-standard-edition my_project_name
```
piotrgradzinski: Wouldn't  be good to add information on how to create a project with a certain version like
`composer create-project symfony/framework-standard-edition my_project_name 2.8`
or
`composer create-project symfony/framework-standard-edition my_project_name 3.4`

fabpot:
IMHO, we should just add a mention at the top of the README mentioning that from 3.4 on, using Flex is better and that Symfony 4.0+ won't be supported by the installer. But for all other versions (that are still supported), the installer is still the recommended tool.

fabpot:
See http://symfony.com/blog/the-end-of-the-symfony-installer for the blog post announcing this.

